PR #152, opened 2020-09-25
EOS-13610 - Mgmt network alert not coming and as a result no failover happened

mariyappanp:
Alerts published by RMQ is not received by CSM due to jsonMsg syntax validation failure.
Event field is set to None. Fix is to mention event field to empty string.


vaibhavparatwar (commented):
@mariyappanp  Have you checked for other sensors also this is not the case?

mariyappanp:
> @mariyappanp Have you checked for other sensors also this is not the case?

Yes, checked over the logs for all other sensor response messages. None other sensor response has None type.

vaibhavparatwar (approved):

indrajitzagade (approved):

thavanathan (approved):
Looks good.

PR #156, opened 2020-09-26
EOS-13684: SSPL sync issue to read Provisioner python API

Madhura-08:
- Modify provisioner console command path
- Add refactoring

## Problem Statement
<pre>
  <code>
    Story Ref (if any): 
    EOS-13684: SSPL sync issue to read Provisioner python API
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Please add Problem decription here...
  </code>
</pre>
## Solution
<pre>
  <code>
    Please add summary of the change,List any dependencies that are required for this change.
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
    - [ ] No
  </code>
</pre>
<pre>
  <code>
  </code>
</pre>


low-level/framework/base/sspl_constants.py
```diff
@@ -116,22 +116,35 @@
...
+SERVER_PER_NODE_KEY = 'servers_per_node'
+STORAGE_TYPE_KEY = 'storage_type'
+SERVER_TYPE_KEY = 'server_type'
+STORAGE_TYPE = 'physical'
+SERVER_TYPE = 'rbod'
+
 
+try:
+    print('fetching the setup values:')
+    process = subprocess.Popen(['sudo', '/usr/local/bin/provisioner', 'get_setup_info'], \
```
indrajitzagade, line 130: During consul vip change we had moved such stuff to a utility file since it is not a good idea to have it in this file. Can we move this code too to a utility file?

low-level/framework/base/sspl_constants.py
```diff
@@ -116,22 +116,35 @@
...
+                shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    _result, _ret_code = process.communicate()[0],process.returncode
+    if _ret_code == 0:
+        setup_info = eval(_result.decode("utf-8").rstrip())
+        STORAGE_TYPE = setup_info[STORAGE_TYPE_KEY].lower()
+        SERVER_TYPE = setup_info[SERVER_TYPE_KEY].lower()
+    logger.info(f'Storage Type: {STORAGE_TYPE}, Server Type: {SERVER_TYPE}')
+    print(f'values: {STORAGE_TYPE} {SERVER_TYPE}')
```
indrajitzagade, line 139: Why print and logger both?

low-level/framework/base/sspl_constants.py
Madhura-08: Right. But problem is, that file gets packed into binary and python path gets set to framework directory which is inside binary. Now, this file gets installed outside binary (and inside binary also I think). And It gets executed at init step. And at that stage, it doesn't find that path which is inside binary. That is why we can't use that Utility class here !

low-level/framework/base/sspl_constants.py
Madhura-08: Because, It gets executed at init level and I think inside binary as well.

low-level/framework/base/sspl_constants.py
```diff
@@ -116,22 +116,35 @@
...
+    logger.debug(f'Considering default values for Storage Type: {STORAGE_TYPE} and Server Type: {SERVER_TYPE}')
 except Exception as err:
-    logger.debug(f"Error in getting setup information of server and storage type : {err}")
-    print(f"Error in getting setup information of server and storage type : {err}")
-    storage_type = 'virtual'
-    server_type = 'virtual'
-    logger.debug(f"Considering default storage type : '{storage_type}'")
-    logger.debug(f"Considering default server type : '{server_type}'")
+    logger.debug(f'Error in getting setup information of server and storage type : {err}')
+    print(f'Error in getting setup information of server and storage type : {err}')
```
thavanathan, line 146: can we use logger.error?

low-level/framework/base/sspl_constants.py
```diff
@@ -116,22 +116,35 @@
...
-    server_type = setup_info['server_type'].lower()
-    logger.info(f"Storage Type : '{storage_type}'")
-    logger.info(f"Server Type '{server_type}'")
+# Provisioner API Call Keys
+NODE_KEY = 'nodes'
+SERVER_PER_NODE_KEY = 'servers_per_node'
+STORAGE_TYPE_KEY = 'storage_type'
+SERVER_TYPE_KEY = 'server_type'
+STORAGE_TYPE = 'physical'
+SERVER_TYPE = 'rbod'
```
thavanathan, line 125: These are default types?

thavanathan (commented):
Minor comments

low-level/framework/base/sspl_constants.py
Madhura-08: Yes. With discussion with Indrajit. I kept the same !

low-level/framework/base/sspl_constants.py
Madhura-08: Sure!

PR #169, opened 2020-10-01
EOS-13821 : Clean up old SAS_PORT_SENSOR_DATA* during post_install script

mohitpathak1:
## Problem Statement
<pre>
  <code>
    Story Ref (if any):
    Clean up old SAS_PORT_SENSOR_DATA* during post_install script
  </code>
</pre>
## Problem Description
<pre>
  <code>
    If the format of SAS_PORT_SENSOR_DATA* changes between two builds, it will lead to errors if SSPL uses old data from persistent storage. So it is a good idea to clean up any old data during post_install script.
  </code>
</pre>
## Solution
<pre>
  <code>
    Clean up old data after each software installation/upgrade from the post_install script.
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
    - [ ] No
  </code>
</pre>
<pre>
  <code>
  </code>
</pre>

### DIFF
--- a/low-level/sspl-ll.spec
+++ b/low-level/sspl-ll.spec
@@ -143,6 +143,7 @@ systemctl stop sspl-ll.service 2> /dev/null
 rm -f /etc/polkit-1/rules.d/sspl-ll_dbus_policy.rules
 rm -f /etc/dbus-1/system.d/sspl-ll_dbus_policy.conf
 [ "$1" == "0" ] && rm -f /opt/seagate/%{product_family}/sspl/sspl_init
+rm -f /var/cortx/sspl/data/server/SAS_PORT_SENSOR_DATA*
 
 %files
 %defattr(-,sspl-ll,root,-)


low-level/files/opt/seagate/sspl/bin/sspl_post_install
```diff
@@ -111,9 +111,12 @@ cp -R $SSPL_BASE_DIR/resources/iem/iec_mapping/* $PRODUCT_BASE_DIR/iem/iec_mappi
...
-# Skip this step if sspl is being configured for node replacement scenario as rabbitmq cluster is 
+# Skip this step if sspl is being configured for node replacement scenario as rabbitmq cluster is
 # already configured on the healthy node
 # Configure rabbitmq
 [ -f $REPLACEMENT_NODE_ENV_VAR_FILE ] || {
     [ "$RMQ_CLUSTER" == true ] && $script_dir/rabbitmq_setup
 }
+
+# CLean up SASPortSensor data from previous installation
+rm -f /var/cortx/sspl/data/server/SAS_PORT_SENSOR_DATA*
```
cdeshmukh, line 122: @mohitpathak1  we need this in  ssp rpm spec file post install step

cdeshmukh (approved):

low-level/sspl-ll.spec
```diff
@@ -94,6 +94,7 @@ chmod 644 $STATE_FILE
 mkdir -p /var/%{product_family}/sspl/bundle /var/log/%{product_family}/sspl /etc/sspl
 SSPL_DIR=/opt/seagate/%{product_family}/sspl
 CFG_DIR=$SSPL_DIR/conf
+rm -f /var/cortx/sspl/data/server/SAS_PORT_SENSOR_DATA*
```
ypise, line 97: @mohitpathak1 rm should ideally go to `%postun`

cdeshmukh: so %post would remove any pre-existing file on pkg installation and %postun would remove file on current pkg uninstallation, and agree latter is cleaner as it wont keep any unwanted files on system. 
However for update case, for older rpms where this change(%postun step)  in not in place, wont execute this cleanup and we would end up hitting same issue, so safer in %post stage only.

@ypise ?

ypise: Right `Update = Uninstall + Install`
Ref: https://www.golinuxhub.com/2018/05/how-to-execute-script-at-pre-post-preun-postun-spec-file-rpm/

mohitpathak1: Moved the change to `%postun`

saurabhkhanvilkar (approved):

low-level/sspl-ll.spec
cdeshmukh: >In post stage, the file is being removed unchecked, so if file is missing, installation would fail.
we have "rm -f" to avoid failing that step, if file exists or not.

Secondly for update case, lets say build 128 -> build 146 failed, now with build 147 we will add this rm -f step under %postun stage of rpm spec file, which would get executed from build 147 rpm uninstallation & stale file would be cleaned up. so all updates from build 147 to any forward build will not have this issue.

But what about update from builds previous to 147, like 128 (or till 146) to 147 (or further build) update, it would still fail, given there is no %postun step which would executed on build <=146, so stale file will remain as it is. To avoid this situation, if we cleanup in %post stage from build 147 onwards, update would pass from any build <147 to build >= 147, given stale file would get cleaned on new rpm installation.

If not clear, lets have a quick call. Ping me when you are available.

cdeshmukh (approved):

low-level/sspl-ll.spec
cdeshmukh: After discussion with Surabh, realize we need this cleanup stage at %postun in case of rollback too (like 147 fails to install, then it should clean up this file). And in %post, for reason what i have mentioned above

cdeshmukh (changes requested):
Need cleanup at both %post & %postun stage

saurabhkhanvilkar (changes requested):
this needs more discussion. me and mohit talked about another scenario which will be a problem, if we make this change. blocking this PR for now, till will discuss and conclude.

mohitpathak1:
Cancelling this PR for now

PR #180, opened 2020-10-05
Changes for EOS-13166 SSPL signal handler fix

ArchanaLimaye:
Modified the SSPL signal handler to disable handling of new SIGHUPs
 while one is being processed

Signed-off-by: Archana Limaye <archana.limaye@seagate.com>

## Problem Statement
<pre>
  <code>
   SSPL signal handler does not ignore additional SIGHUPs while processing one SIGHUP
  </code>
</pre>
## Problem Description
<pre>
  <code>
    SSPL signal handler does not ignore additional SIGHUPs while processing one SIGHUP
  </code>
</pre>
## Solution
<pre>
  <code>
    Modified the SSPL code so that any additional SIGHUPs while one is being processed are ignored
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - Yes
  </code>
</pre>
<pre>
  <code>
 Deployed SSPL on VM.
 Issued multiple SIGHUPs.
 Ensured that only one SIGHUP is being handled and others are ignored; by checking the logs 
  </code>
</pre>

### DIFF
--- a/low-level/framework/sspl_ll_d
+++ b/low-level/framework/sspl_ll_d
@@ -613,6 +613,11 @@ def signal_handler(signal_number, frame):
        The entries in text file should be in form of <key=value>.
     """
     logger.debug("signal_handler called with {0}".format(signal_number))
+
+    # Ignore a new  SIGHUP while handling current SIGHUP
+    # Signal will be enabled at the end of this handler
+    signal.signal(signal.SIGHUP, signal.SIG_IGN)
+
     entries = dict()
     state = "active"
     global sspl_role_state
@@ -655,6 +660,9 @@ def signal_handler(signal_number, frame):
         # thread is not yet ready,save the request to serve it later when thread
         # will be ready.
         sspl_role_state = state
+
+    # Enable the signal handler 
+    signal.signal(signal.SIGHUP, signal_handler)
 
 def send_thread_controller_request(state):
     '''Creates a internal request for thread controller
--- a/low-level/framework/sspl_ll_d
+++ b/low-level/framework/sspl_ll_d
@@ -618,10 +618,10 @@ def signal_handler(signal_number, frame):
 
     logger.debug("signal_handler called with {0}".format(signal_number))
 
-    entries = dict()
-    state = "active"
-    global sspl_role_state
     try:
+        entries = dict()
+        state = "active"
+        global sspl_role_state
         with open(STATE_FILE) as state_file:
 
             for line in state_file.readlines():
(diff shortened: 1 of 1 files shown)


saurabhkhanvilkar (approved):

low-level/framework/sspl_ll_d
```diff
@@ -613,6 +613,11 @@ def signal_handler(signal_number, frame):
        The entries in text file should be in form of <key=value>.
     """
     logger.debug("signal_handler called with {0}".format(signal_number))
+
+    # Ignore a new  SIGHUP while handling current SIGHUP
+    # Signal will be enabled at the end of this handler
+    signal.signal(signal.SIGHUP, signal.SIG_IGN)
```
indrajitzagade, line 619: I would make this a first statement to be executed after we enter the signal handler. Was there any issue seen in that?

ArchanaLimaye: Yes we can do that. 
FYI I never saw the log above i.e. logger.debug("signal_handler called with {0}".format(signal_number))  
coming up on VM in my testing ....
Its not visible in the quick experiment that was done either...
Any idea why that might be ?

indrajitzagade: Did you check the logging level?

ArchanaLimaye: Did not check ; but though that since 'info' is getting displayed; 'debug' should get displayed to; info is the lowest level rt ? 
Nyways will move the statement

cdeshmukh: > /opt/seagate/cortx/hare/bin/consul kv put sspl/config/SYSTEM_INFORMATION/log_level DEBUG

@ArchanaLimaye change to DEBUG log level and you should see that log. And if this is working fine, then you should see this log only once.

@indrajitzagade agree, though barring logger.debug line its first line only. To check i think have log before signal is fine, or we would never know while testing or in production that signal handler indeed ignoring second signal, while first signal is still under processing.

cdeshmukh (approved):

low-level/framework/sspl_ll_d
saurabhkhanvilkar: i think debug is the lowest level.

thavanathan: Instead of this, can we use a global lock variable that will be set to True when handling the signal? If we change the signal handler itself, we may never know that a signal is received or not in the middle of handling another signal.

ArchanaLimaye: @cdeshmukh I moved the SIG_IGN to before log message based on Indrajit's comment.....
should this be reverted back ? 

@thavanathan you mean some variable in consul or file ? 
>>we may never know that a signal is received or not in the middle of handling another signal.
Also do we need to be aware if the same signal is received again ?  SIG_IGN applies only to SIGHUP

indrajitzagade (approved):

low-level/framework/sspl_ll_d
cdeshmukh: this handler is for catching SIGHUP, so that is enough for this one.

cdeshmukh (approved):

thavanathan (approved):

PR #195, opened 2020-10-08
EOS-13850 5U Disk fault alerts not seen after consul disable and enable

VijayThakkar132:
Signed-off-by: Vijay Thakkar <vijaykumar.thakkar@seagate.com>

## Problem Statement
<pre>
  <code>
    Story Ref (if any):EOS-13850 
    5U Disk fault alerts not seen after "pcs resource disable/enable consul-c1"
  </code>
</pre>
## Problem Description
<pre>
  <code>
    5U disks alerts not seen after disabling and enabling again consul agent
  </code>
</pre>
## Solution
<pre>
  <code>
    1. Data corruption was happening in the put method of consul when consul was disable and suddenly comes back during the consul connection retires. Fixed issue
    2. Also added handing to skip alerts when consul connection is down.
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
  </code>
</pre>
<pre>
  <code>
    Validated rpms on the hardware setup and observed that alerts are coming as per the expectation.
  </code>
</pre>

### DIFF
--- a/low-level/framework/utils/consulstore.py
+++ b/low-level/framework/utils/consulstore.py
@@ -76,7 +76,7 @@ class ConsulStore(Store):
                 break
 
     def _consul_get(self, key, **kwargs):
-        """Load consul data from the given key"""
+        """Load consul data from the given key."""
         data = None
         status = "Failure"
 
--- a/low-level/framework/utils/consulstore.py
+++ b/low-level/framework/utils/consulstore.py
@@ -83,7 +83,7 @@ class ConsulStore(Store):
         for retry_index in range(0, MAX_CONSUL_RETRY):
             try:
                 key = self._get_key(key)
-                data = self.consul_conn.kv.get(key)[1]
+                data = self.consul_conn.kv.get(key, recurse=_opt_recurse)[1]
                 if data:
                     data = data["Value"]
                     try:
--- a/low-level/framework/utils/consulstore.py
+++ b/low-level/framework/utils/consulstore.py
@@ -56,12 +56,13 @@ class ConsulStore(Store):
 
     def put(self, value, key, pickled=True):
         """ write data to given key"""
+        key = self._get_key(key)
+        if pickled:
+            value = pickle.dumps(value)
+
         for retry_index in range(0, MAX_CONSUL_RETRY):
             try:
-                key = self._get_key(key)
-                if pickled:
-                    pickled_value = pickle.dumps(value)
-                self.consul_conn.kv.put(key, pickled_value)
+                self.consul_conn.kv.put(key, value)
                 break
 
             except requests.exceptions.ConnectionError as connerr:
@@ -74,39 +75,10 @@ class ConsulStore(Store):
                     .format(gerr, key))
                 break
 
-    def get(self, key, **kwargs):
-        """ Load data from given key"""
-        data = None
-        for retry_index in range(0, MAX_CONSUL_RETRY):
-            try:
-                _opt_recurse = kwargs.get("recurse", False)
-                key = self._get_key(key)
-                data = self.consul_conn.kv.get(key, recurse=_opt_recurse)[1]
-                if data:
-                    data = data["Value"]
-                    try:
-                        data = pickle.loads(data)
-                    except:
-                        pass
-                break
-
-            except requests.exceptions.ConnectionError as connerr:
-                logger.warn("Error[{0}] consul connection refused Retry Index {1}" \
-                    .format(connerr, retry_index))
-                time.sleep(WAIT_BEFORE_RETRY)
-
-            except Exception as gerr:
-                logger.warn("Error[{0}] while reading data from consul {1}" \
-                    .format(gerr, key))
-                break
-        return data
-
-    def exists(self, key):
-        """check if key exists
-        """
+    def _consul_get(self, key, **kwargs):
+        """Load consul data from the given key"""
         data = None
         status = "Failure"
-        is_exists = False
 
         for retry_index in range(0, MAX_CONSUL_RETRY):
             try:
@@ -131,10 +103,23 @@ class ConsulStore(Store):
                     .format(gerr, key))
                 break
 
-        if data is not None:
-            is_exists = True
+        return data, status
 
-        return is_exists, status
+    def get(self, key, **kwargs):
+        """ Load data from given key"""
+        data, _ = self._consul_get(key, **kwargs)
+        return data
+
+    def exists(self, key):
+        """check if key exists
+        """
+        key_present = False
+        data, status = self._consul_get(key)
+
+        if data is not None:
+            key_present = True
+
+        return key_present, status
 
     def delete(self, key):
         """ delete a key
--- a/low-level/framework/utils/consulstore.py
+++ b/low-level/framework/utils/consulstore.py
@@ -82,6 +82,7 @@ class ConsulStore(Store):
 
         for retry_index in range(0, MAX_CONSUL_RETRY):
             try:
+                _opt_recurse = kwargs.get("recurse", False)
                 key = self._get_key(key)
                 data = self.consul_conn.kv.get(key, recurse=_opt_recurse)[1]
                 if data:
--- a/low-level/framework/utils/consulstore.py
+++ b/low-level/framework/utils/consulstore.py
@@ -60,8 +60,8 @@ class ConsulStore(Store):
             try:
                 key = self._get_key(key)
                 if pickled:
-                    value = pickle.dumps(value)
-                self.consul_conn.kv.put(key, value)
+                    pickled_value = pickle.dumps(value)
+                self.consul_conn.kv.put(key, pickled_value)
                 break
 
             except requests.exceptions.ConnectionError as connerr:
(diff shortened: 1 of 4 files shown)


low-level/framework/utils/consulstore.py
```diff
@@ -60,8 +60,8 @@ def put(self, value, key, pickled=True):
             try:
                 key = self._get_key(key)
                 if pickled:
-                    value = pickle.dumps(value)
```
VijayThakkar132, line 63: Observed that storing pickled value in the same variable causing data corruption when consul is down and if suddenly it comes back that time it will pickle value again which is already pickled.

low-level/framework/utils/consulstore.py
```diff
@@ -104,10 +104,37 @@ def get(self, key, **kwargs):
     def exists(self, key):
         """check if key exists
         """
-        if self.get(key):
-            return True
-        else:
-            return False
+        data = None
```
VijayThakkar132, line 107: We don't have any way to differentiate exists method has returns None due to connection error or path not exists.
And Internally it uses Get method of the ConsulStore. I re-wrote exists API to avoid changes in the Get method at this stage as it used more then 20 files. We can improve later and add error handling in the all the API of the ConsulStore.

sandeepsgt (commented):
Do we need to skip alerts from other sensors(which are using consul stor) also ?

low-level/sensors/impl/platforms/realstor/realstor_disk_sensor.py
```diff
@@ -146,11 +147,15 @@ def run(self):
...
+            if not self.invalidate_latest_disks_info:
+                # Polling system status
+                self.rssencl.get_system_status()
 
-            # check for disk faults & raise if found
-            self._rss_check_disk_faults()
+                # check for disk faults & raise if found
+                self._rss_check_disk_faults()
+            else:
+                logger.warn("Ignore disk faults check due to consul error")
```
cdeshmukh, line 158: same comment as line 150

Comment can be more to the point
"Can not validate disk faults or presence due to persistence store error"

VijayThakkar132:
I checked all other enclosure related sensors and I could see we always use local copy to get previous status of the sensor. We also store current status of the sensor in the Consul after sending alert to the rabbitmq but it get used in the following cases which are bit rare.

1. Get data from the consul during initialization.
2. Get data from the consul in case of timeout.

low-level/sensors/impl/platforms/realstor/realstor_disk_sensor.py
```diff
@@ -333,19 +340,31 @@ def rss_cliapi_poll_disks(self, disk):
...
+                                if prevsn != sn or prevhealth != health:
+                                    # Rename path
+                                    store.put(store.get(dcache_path), dcache_path + ".prev")
+                                    store.delete(dcache_path)
+
+                                    store.put(drive, dcache_path)
+                        elif not path_exists and ret_val == "Success":
                             store.put(drive, dcache_path)
+                        else:
+                            # Invalidate latest disks info if consul connection error is found
```
cdeshmukh, line 360: Invalidate latest disks info if persistence store error encountered

cdeshmukh (changes requested):

low-level/framework/utils/consulstore.py
indrajitzagade: Good catch @VijayThakkar132

low-level/framework/utils/consulstore.py
indrajitzagade: Agree

cdeshmukh:
> I checked all other enclosure related sensors and I could see we always use local copy to get previous status of the sensor. 
Check for node data sensors too, where usage threshold for cpu/memory/disk space are kept in consul store. This threshold should be read during init time only and during every iteration, so once in memory sspl would be good to raise alerts for these anyways. IF during sspl/sensor threads init itself consul is seen down, then need to keep retrying to read these thresholds until consul is up, until that corresponding alerts cant be expected again.

> We also store current status of the sensor in the Consul after sending alert to the rabbitmq but it get used in the following cases which are bit rare.

> 1. Get data from the consul during initialization.
Why above case is rare? sspl init or node re/boot will always exercise this path

> 2. Get data from the consul in case of timeout.
This is the case we are handling i believe here.

VijayThakkar132:
> I checked all other enclosure related sensors and I could see we always use local copy to get previous status of the sensor.
Check for node data sensors too, where usage threshold for cpu/memory/disk space are kept in consul store. This threshold should be read during init time only and during every iteration, so once in memory sspl would be good to raise alerts for these anyways. IF during sspl/sensor threads init itself consul is seen down, then need to keep retrying to read these thresholds until consul is up, until that corresponding alerts cant be expected again.

I agree with @cdeshmukh. We should not proceed further till consul is up and this exercise is required for mostly all the sensors. Can we create separate jira for the same otherwise this task get delayed. @indrajitzagade your thought please.

low-level/framework/utils/filestore.py
```diff
@@ -131,7 +131,7 @@ def _load_json_file(self, key):
     def exists(self, key):
         """check if key exists
         """
-        return os.path.exists(key)
+        return os.path.exists(key), "Success"
```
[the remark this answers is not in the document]
VijayThakkar132: done

low-level/sensors/impl/platforms/realstor/realstor_disk_sensor.py
```diff
@@ -146,11 +147,15 @@ def run(self):
             # insertion/removal detected
             self._rss_check_disks_presence()
 
-            # Polling system status
-            self.rssencl.get_system_status()
+            #Do not proceed further if latest disks info is not valid due to consul connection error
```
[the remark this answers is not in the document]
VijayThakkar132: done

low-level/sensors/impl/platforms/realstor/realstor_disk_sensor.py
VijayThakkar132: done

low-level/sensors/impl/platforms/realstor/realstor_disk_sensor.py
VijayThakkar132: done

low-level/framework/utils/consulstore.py
```diff
@@ -74,20 +75,22 @@ def put(self, value, key, pickled=True):
...
+    def _consul_get(self, key, **kwargs):
+        """Load consul data from the given key."""
         data = None
+        status = "Failure"
+
         for retry_index in range(0, MAX_CONSUL_RETRY):
             try:
-                _opt_recurse = kwargs.get("recurse", False)
                 key = self._get_key(key)
-                data = self.consul_conn.kv.get(key, recurse=_opt_recurse)[1]
```
cdeshmukh, line 84: why removing the second argument here ? Its currently serving get() code which has very wide impact, have you thought of all scenarios ?

low-level/framework/utils/filestore.py
```diff
@@ -129,9 +129,17 @@ def _load_json_file(self, key):
...
+        """check if key is present
         """
-        return os.path.exists(key)
+        key_present = False
+        status = "Failure"
+        try:
+            key_present = os.path.exists(key)
+            status = "Success"
+        except Exception as gerr:
+            logger.warn("Error{0} while checking if key is present".format(gerr))
```
cdeshmukh, line 140: logger.warn("Error while checking if {0} is present".format(gerr))

cdeshmukh (changes requested):
Also no other sensor changes needed ?

low-level/framework/utils/consulstore.py
VijayThakkar132: By mistake, I removed second argument in last commit (copy-paste mistake). Thanks for pointing out.

VijayThakkar132: Corrected

low-level/framework/utils/filestore.py
VijayThakkar132: done

VijayThakkar132:
I prepared list of sensors which are under sensors/impl/generic 

1. CPU sensor - Using filestore
2. IEM sensor - Not using persistence store
3. node_data - Not using persistence store
4. node_hw - Using filestore
5. node_memory_fault - Using filestore
6. node_sas_port - Using filestore
7. raid - Not using persistence store
8. raid_integrity_data - Not using persistence store
9. SMR_drive_data - Not using persistence store
10. SNMP_traps - Not using persistence store

@cdeshmukh @indrajitzagade

VijayThakkar132:
Actually, store_queue and resource_health_view also use consul store as a persistence store. Need to analyze impact when consul is down.

cdeshmukh:
> node_data - Not using persistence store

Threshold is coming from config directly then

> 9. SMR_drive_data - Not using persistence store
> 10. SNMP_traps - Not using persistence store

These are not in use btw !

cdeshmukh:
> store_queue

cdeshmukh:
> Actually, store_queue and resource_health_view also use consul store as a persistence store. Need to analyze impact when consul is down.

rabbitmq store_queue uses consul to accumulate messages when rabbitmq is down, atlast ges through consulstore get / put calls. And i dont see return value being checked there. Better run through that flow once.

resource_health_view creates first time persistence cache for sspl if one doesnt exists. If consul is down during this time, its ok healthview scripts should just move on. When sspl boots up, will take care of creating persistence cache is none exists, again when consul is up though.

VijayThakkar132:
From the code, It looks like if both rabbitmq and consul goes down at same time then we may loss alerts and also ended up by incrementing incorrect counter because there is no check whether store.put(From StoreQueue class)succeeded or not. Same applies to store.get also. @cdeshmukh

cdeshmukh:
> > I checked all other enclosure related sensors and I could see we always use local copy to get previous status of the sensor.
> > Check for node data sensors too, where usage threshold for cpu/memory/disk space are kept in consul store. This threshold should be read during init time only and during every iteration, so once in memory sspl would be good to raise alerts for these anyways. IF during sspl/sensor threads init itself consul is seen down, then need to keep retrying to read these thresholds until consul is up, until that corresponding alerts cant be expected again.
> 
> I agree with @cdeshmukh. We should not proceed further till consul is up and this exercise is required for mostly all the sensors. Can we create separate jira for the same otherwise this task get delayed. @indrajitzagade your thought please.

Yes this is what we discussed with Ujjwal that day right. Such things needs design changes and of test cycles, we will plan for LDR R2.

VijayThakkar132:
Agree

cdeshmukh:
> From the code, It looks like if both rabbitmq and consul goes down at same time then we may loss alerts and also ended up by incrementing incorrect counter because there is no check whether store.put(From StoreQueue class)succeeded or not. Same applies to store.get also. @cdeshmukh

Can you create a jira with all these intricate details for LDR R2, we will take care during revamp of this solution.

cdeshmukh (approved):

sandeepsgt (approved):

saurabhkhanvilkar (approved):

thavanathan (approved):

PR #224, opened 2020-10-26
SSPL Test to use actual RMQ credentials

mohitpathak1:
## Problem Statement
<pre>
  <code>
    Story Ref (if any):
    EOS-14597
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Sanity test fails because of RMQ connection issue
  </code>
</pre>
## Solution
<pre>
  <code>
    Use actual credential values instead of default
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
    - [ ] No
  </code>
</pre>
<pre>
  <code>
  </code>
</pre>

### DIFF
--- a/sspl_test/framework/utils/encryptor.py
+++ b/sspl_test/framework/utils/encryptor.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License along
+# with this program. If not, see <https://www.gnu.org/licenses/>. For any questions
+# about this software or licensing, please email opensource@seagate.com or
+# cortx-questions@seagate.com.
+
+
+#****************************************************************************
+# Description:       Common utility functions required for encryption purpose
+#****************************************************************************
+# TODO - Avoid duplicate code between sspl and sspl_test
+
+from cortx.utils.security.cipher import Cipher
+
+def gen_key(cluster_id, service_name):
+    # Generate key for decryption
+    key = Cipher.generate_key(cluster_id, service_name)
+    return key
+
+
+def encrypt(key, text):
+    """Encrypt sensitive data. Ex: RabbitMQ credentials."""
+    # Before encrypting text we need to convert string to bytes using encode()
+    # method
+    return Cipher.encrypt(key, text.encode())
+
+
+def decrypt(key, text, caller=None):
+    """Decrypt the <text>."""
+    decrypt_text = Cipher.decrypt(key, text).decode()
+    return decrypt_text
--- a/sspl_test/rabbitmq/rabbitmq_egress_processor.py
+++ b/sspl_test/rabbitmq/rabbitmq_egress_processor.py
@@ -26,6 +26,8 @@ import time
 from sspl_test.framework.base.module_thread import ScheduledModuleThread
 from sspl_test.framework.base.internal_msgQ import InternalMsgQ
 from sspl_test.framework.utils.service_logging import logger
+from sspl_test.framework.utils import encryptor
+from sspl_test.framework.base.sspl_constants import ServiceTypes
 from .rabbitmq_sspl_test_connector import RabbitMQSafeConnection
 
 import ctypes
@@ -213,12 +215,16 @@ class RabbitMQegressProcessor(ScheduledModuleThread, InternalMsgQ):
             self._site_id = self._conf_reader._get_value_with_default(
                 self.SYSTEM_INFORMATION, self.SITE_ID, '')
 
+            # Decrypt RabbitMQ Password
+            decryption_key = encryptor.gen_key(self._cluster_id, ServiceTypes.RABBITMQ.value)
+            self._password = encryptor.decrypt(decryption_key, self._password.encode('ascii'), "RabbitMQegressProcessor")
 
             if self._iem_route_addr != "":
                 logger.info("         Routing IEMs to host: %s" % self._iem_route_addr)
                 logger.info("         Using IEM exchange: %s" % self._iem_route_exchange_name)
         except Exception as ex:
             logger.exception("RabbitMQegressProcessor, _read_config: %r" % ex)
+            raise
 
     def _add_signature(self):
         """Adds the authentication signature to the message"""
--- a/sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
+++ b/sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
@@ -33,6 +33,8 @@ from sspl_test.framework.base.module_thread import ScheduledModuleThread
 from sspl_test.framework.base.internal_msgQ import InternalMsgQ
 from sspl_test.framework.utils.service_logging import logger
 from sspl_test.framework.base.sspl_constants import RESOURCE_PATH
+from sspl_test.framework.utils import encryptor
+from sspl_test.framework.base.sspl_constants import ServiceTypes
 from .rabbitmq_sspl_test_connector import RabbitMQSafeConnection
 import ctypes
 SSPL_SEC = ctypes.cdll.LoadLibrary('libsspl_sec.so.0')
@@ -53,6 +55,9 @@ class RabbitMQingressProcessorTests(ScheduledModuleThread, InternalMsgQ):
     VIRT_HOST             = 'virtual_host'
     USER_NAME             = 'username'
     PASSWORD              = 'password'
+
+    SYSTEM_INFORMATION_KEY = "SYSTEM_INFORMATION"
+    CLUSTER_ID_KEY = "cluster_id"
 
     JSON_ACTUATOR_SCHEMA = "SSPL-LL_Actuator_Response.json"
     JSON_SENSOR_SCHEMA   = "SSPL-LL_Sensor_Response.json"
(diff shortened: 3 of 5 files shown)


sspl_test/run_qa_test.sh
```diff
@@ -48,6 +48,11 @@ flask_help()
 
 pre_requisites()
 {
+    # copy RMQ password to sspl_test/config
+    pw=$($CONSUL_PATH/consul kv get sspl/config/RABBITMQINGRESSPROCESSOR/password)
```
mohitpathak1, line 52: @cdeshmukh @thavanathan please confirm if it is okay to assume that all 4 processors will have a common password

sspl_test/framework/utils/encryptor.py
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
thavanathan, line 1: Can we use `framework.utils.encryptor`? If encryption logic changes in future, we need to change here again.

sspl_test/framework/utils/encryptor.py
```diff
@@ -0,0 +1,46 @@
...
+
+def decrypt(key, text, caller=None):
+    '''Decrypt the <text>'''
+    decrypt_text = text
+    try:
+        decrypt_text = Cipher.decrypt(key, text).decode()
+        return decrypt_text
+    except CipherInvalidToken as e:
+        logger.error("{0}:Password decryption failed requested by {1}.".format(e, caller))
+        return decrypt_text.decode()
```
thavanathan, line 46: Please raise the exception. There is no use in using encrypted string.

sspl_test/rabbitmq/rabbitmq_egress_processor.py
```diff
@@ -213,6 +215,9 @@ def _read_config(self):
             self._site_id = self._conf_reader._get_value_with_default(
                 self.SYSTEM_INFORMATION, self.SITE_ID, '')
 
+            # Decrypt RabbitMQ Password
+            decryption_key = encryptor.gen_key(self._cluster_id, ServiceTypes.RABBITMQ.value)
```
thavanathan, line 219: Can you please check where are we reading self._cluster_id from?

sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
```diff
@@ -54,6 +56,9 @@ class RabbitMQingressProcessorTests(ScheduledModuleThread, InternalMsgQ):
     USER_NAME             = 'username'
     PASSWORD              = 'password'
 
+    SYSTEM_INFORMATION = "SYSTEM_INFORMATION"
+    CLUSTER_ID = "cluster_id"
```
thavanathan, line 60: CLUSTER_ID_KEY

sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
```diff
@@ -54,6 +56,9 @@ class RabbitMQingressProcessorTests(ScheduledModuleThread, InternalMsgQ):
     USER_NAME             = 'username'
     PASSWORD              = 'password'
 
+    SYSTEM_INFORMATION = "SYSTEM_INFORMATION"
```
thavanathan, line 59: SYSTEM_INFORMATION_KEY

sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
```diff
@@ -231,6 +236,11 @@ def _configure_exchange(self):
                 self._routing_key,
                 self._queue_name
             )
+            cluster_id = self._conf_reader._get_value_with_default(
```
thavanathan, line 239: let's assign it to self.cluster_id

sspl_test/run_qa_test.sh
thavanathan: Yes. But to be safe, let's read from respective processor keys from consul.

thavanathan (changes requested):

sspl_test/framework/utils/encryptor.py
cdeshmukh: agree, we need to reuse sspl code here

sspl_test/rabbitmq/rabbitmq_egress_processor.py
cdeshmukh: line 212, from common config only, which looks fine to me.

sspl_test/run_qa_test.sh
cdeshmukh: till last week they were same, now may not, after Provisioner possibly generating separate password for different processor, so better read & use separately

cdeshmukh (changes requested):

sspl_test/framework/utils/encryptor.py
mohitpathak1: If we use `framework.utils.encryptor` inside SSPL Test, it leads to an import error, hence created this new file. If we want to use the same file, we will need to figure out a way to copy it to sspl_test while building the rpms.

sspl_test/run_qa_test.sh
mohitpathak1: Done

sspl_test/framework/utils/encryptor.py
cdeshmukh: for same reason sspl_constants file copy is also created i believe. We need to overcome this or we will keep replicating code & bloat so much. 
Adjusting python path should help ideally, but seems you would have tried already, may not be that straight forward.

thavanathan: Yeah. Let's keep it for now.

cdeshmukh: Alright, keep a TODO note & create a task ticket then to address in future. both for this one & sspl-constants & alike code.

cdeshmukh (approved):

sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
```diff
@@ -231,6 +236,15 @@ def _configure_exchange(self):
...
                 self._queue_name
             )
+            self.cluster_id = self._conf_reader._get_value_with_default(
+                self.SYSTEM_INFORMATION_KEY, self.CLUSTER_ID_KEY, '')
+            # Decrypt RabbitMQ Password
+            decryption_key = encryptor.gen_key(self.cluster_id, ServiceTypes.RABBITMQ.value)
+            try:
+                self._password = encryptor.decrypt(decryption_key, self._password.encode('ascii'), "RabbitMQingressProcessor")
+            except Exception as e:
+                self._password = "sspl4ever"
```
thavanathan, line 246: This would never work :) It is not a default password anymore. Just raise the exception back. Then we would immediately know something went wrong.

sspl_test/rabbitmq/rabbitmq_egress_processor.py
```diff
@@ -213,7 +215,13 @@ def _read_config(self):
             self._site_id = self._conf_reader._get_value_with_default(
                 self.SYSTEM_INFORMATION, self.SITE_ID, '')
 
-
+            # Decrypt RabbitMQ Password
+            decryption_key = encryptor.gen_key(self._cluster_id, ServiceTypes.RABBITMQ.value)
+            try:
+                self._password = encryptor.decrypt(decryption_key, self._password.encode('ascii'), "RabbitMQegressProcessor")
+            except Exception as e:
+                self._password = "sspl4ever"
```
thavanathan, line 223: It is not a default password anymore. Just raise the exception back. Then we would immediately know something went wrong.

sspl_test/rabbitmq/rabbitmq_egress_processor.py
```diff
@@ -27,6 +27,8 @@
 from sspl_test.framework.base.internal_msgQ import InternalMsgQ
 from sspl_test.framework.utils.service_logging import logger
 from .rabbitmq_sspl_test_connector import RabbitMQSafeConnection
+from sspl_test.framework.utils import encryptor
+from sspl_test.framework.base.sspl_constants import ServiceTypes
```
thavanathan, line 31: Let's move these two lines above line 29. Let's keep the import order.

sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
```diff
@@ -36,6 +36,8 @@
 from .rabbitmq_sspl_test_connector import RabbitMQSafeConnection
 import ctypes
 SSPL_SEC = ctypes.cdll.LoadLibrary('libsspl_sec.so.0')
+from sspl_test.framework.utils import encryptor
+from sspl_test.framework.base.sspl_constants import ServiceTypes
```
thavanathan, line 40: Let's move these two lines above line 36 to keep the import order.

thavanathan (changes requested):

sspl_test/rabbitmq/rabbitmq_ingress_processor_tests.py
```diff
@@ -231,6 +236,11 @@ def _configure_exchange(self):
                 self._routing_key,
                 self._queue_name
             )
+            self.cluster_id = self._conf_reader._get_value_with_default(
+                self.SYSTEM_INFORMATION_KEY, self.CLUSTER_ID_KEY, '')
+            # Decrypt RabbitMQ Password
+            decryption_key = encryptor.gen_key(self.cluster_id, ServiceTypes.RABBITMQ.value)
+            self._password = encryptor.decrypt(decryption_key, self._password.encode('ascii'), "RabbitMQingressProcessor")
         except Exception as ex:
             logger.exception("_configure_exchange: %r" % ex)
```
thavanathan, line 246: add a line here at 246 with just one below stmt
`raise`

thavanathan (changes requested):

thavanathan (approved):
Looks good.

sspl_test/framework/utils/encryptor.py
ujjwalpl: We need to avoid "do later" approach because its piling up with more TODOs. Technically we have the library is available in cortx-py-utils and can call directly instead of duplicating it.

PR #231, opened 2020-10-30
Rabbitmq upgrade

Madhura-08:
## Problem Statement
<pre>
  <code>
    Story Ref (if any):
   Make SSPL compatiblw with latest RabbitMQ stable version
  </code>
</pre>
## Problem Description
<pre>
  <code>
    RabbitMQ configuration script was failing for RabbitMQ version 3.8.9. some assert statements were failing which was version specific. 
 </code>
</pre>
## Solution
<pre>
  <code>
    Foe now, removed that assert.
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
     </code>
</pre>
<pre>
  <code>
    Install RabbitMQ 3.8.9 with its required dependency like erlang >= 22.x
    Install and run SSPL
    Perform SSPL sanity
  </code>
</pre>

### DIFF
--- a/low-level/files/opt/seagate/sspl/bin/sspl_config
+++ b/low-level/files/opt/seagate/sspl/bin/sspl_config
@@ -200,13 +200,23 @@ rabbitmq_minor_release=$(echo "${rabbitmq_version::-1}" | sed 's/.$//')
 # Get the Maitenance release version parsed from minor release. (Eg: 8 from 3.8)
 rabbitmq_maintenance_release=$(echo "${rabbitmq_minor_release:2:3}")
 
-# Skip this step if sspl is being configured for node replacement scenario as consul data is already
-# available on healthy node
-# Updating RabbitMQ cluster nodes.
-# In node replacement scenario, avoiding feeding again to avoid over writing already configured values
-# with which rabbitmq cluster may have been created
-[ -f $REPLACEMENT_NODE_ENV_VAR_FILE ] || {
-    # Update cluster_nodes key in consul
+get_cluster_running_nodes() {
+    pout=""
+    case $1 in
+
+    "rabbitmq" )
+        pout=$( get_rabbitmq_cluster_nodes )
+        echo "$pout"
+        ;;
+
+    * )
+        usage
+        ;;
+    esac
+
+}
+
+get_rabbitmq_cluster_nodes() {
     if [ $rabbitmq_major_release -eq 3 ] && [ $rabbitmq_maintenance_release -eq 8 ];
     then
         rabbitmq_cluster_status=$(/usr/sbin/rabbitmqctl cluster_status --formatter json)
@@ -220,7 +230,22 @@ rabbitmq_maintenance_release=$(echo "${rabbitmq_minor_release:2:3}")
         echo "This RabbitMQ version: $rabbitmq_version is not supported"
         exit 1
     fi
+    echo "$pout"
+}
 
+# Skip this step if sspl is being configured for node replacement scenario as consul data is already
+# available on healthy node
+# Updating RabbitMQ cluster nodes.
+# In node replacement scenario, avoiding feeding again to avoid over writing already configured values
+# with which rabbitmq cluster may have been created
+[ -f $REPLACEMENT_NODE_ENV_VAR_FILE ] || {
+
+    message_broker="rabbitmq"
+
+    # Get the running nodes from a cluster
+    pout=$( get_cluster_running_nodes "$message_broker")
+
+    # Update cluster_nodes key in consul
     $CONSUL_PATH/consul kv put sspl/config/RABBITMQCLUSTER/cluster_nodes $pout
     [ -n "$rmq_cluster_nodes" ] && $CONSUL_PATH/consul kv put sspl/config/RABBITMQCLUSTER/cluster_nodes $rmq_cluster_nodes
 }
--- a/low-level/files/opt/seagate/sspl/bin/sspl_config
+++ b/low-level/files/opt/seagate/sspl/bin/sspl_config
@@ -188,6 +188,11 @@ case $cmd in
         ;;
 esac
 
+rabbitmq_version=$(rpm -qi rabbitmq-server | awk -F': ' '/Version/ {print $2}')
+rabbitmq_base_version=$(echo "${rabbitmq_version:0:-3}" | sed 's/.$//')
+rabbitmq_updated_version=$(echo "${rabbitmq_version::-1}" | sed 's/.$//')
+rabbitmq_middle_version=$(echo "${rabbitmq_updated_version:2:3}")
+
 # Skip this step if sspl is being configured for node replacement scenario as consul data is already
 # available on healthy node
 # Updating RabbitMQ cluster nodes.
@@ -195,8 +200,16 @@ esac
 # with which rabbitmq cluster may have been created
 [ -f $REPLACEMENT_NODE_ENV_VAR_FILE ] || {
     # Update cluster_nodes key in consul
-    out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
-    pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
+    if [ $rabbitmq_base_version -eq 3 ] && [ $rabbitmq_middle_version -ge 8 ];
+    then
+        rabbitmq_cluster_status=$(/usr/sbin/rabbitmqctl cluster_status --formatter json)
+        cluster_running_nodes=$(python3 -c "rabbitmq_status=$rabbitmq_cluster_status; print(rabbitmq_status['running_nodes'])")
+        pout=$(echo $cluster_running_nodes | sed 's/rabbit@//g' | cut -d '[' -f2 | cut -d ']' -f1 | tr -d \'\" | tr -d ' ')
+    else
+        out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
+        pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
+    fi
+
     $CONSUL_PATH/consul kv put sspl/config/RABBITMQCLUSTER/cluster_nodes $pout
     [ -n "$rmq_cluster_nodes" ] && $CONSUL_PATH/consul kv put sspl/config/RABBITMQCLUSTER/cluster_nodes $rmq_cluster_nodes
 }
--- a/low-level/files/opt/seagate/sspl/bin/sspl_config
+++ b/low-level/files/opt/seagate/sspl/bin/sspl_config
@@ -217,7 +217,7 @@ rabbitmq_maintenance_release=$(echo "${rabbitmq_minor_release:2:3}")
         out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
         pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
     else
-        echo "Not a supported version"
+        echo "This RabbitMQ version: $rabbitmq_version is not supported"
         exit 1
     fi
 
--- a/low-level/files/opt/seagate/sspl/bin/sspl_config
+++ b/low-level/files/opt/seagate/sspl/bin/sspl_config
@@ -212,9 +212,13 @@ rabbitmq_maintenance_release=$(echo "${rabbitmq_minor_release:2:3}")
         rabbitmq_cluster_status=$(/usr/sbin/rabbitmqctl cluster_status --formatter json)
         cluster_running_nodes=$(python3 -c "rabbitmq_status=$rabbitmq_cluster_status; print(rabbitmq_status['running_nodes'])")
         pout=$(echo $cluster_running_nodes | sed 's/rabbit@//g' | cut -d '[' -f2 | cut -d ']' -f1 | tr -d \'\" | tr -d ' ')
-    else
+    elif [ $rabbitmq_version == "3.3.5" ];
+    then
         out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
         pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
+    else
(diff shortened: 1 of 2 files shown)


low-level/framework/sspl_rabbitmq_reinit
```diff
@@ -270,9 +270,9 @@ def _create_vhost_if_necessary(virtual_host, product):
         vhosts = subprocess.check_output(
             [RABBITMQCTL, 'list_vhosts']
             ).decode("utf-8").split('\n')
-        assert vhosts[0] == 'Listing vhosts ...'
-        assert vhosts[-2] == '...done.'
-        assert vhosts[-1] == ''
+        # assert vhosts[0] == 'Listing vhosts ...'
```
Madhura-08, line 273: These assert statements checks output which is very much version specific. I don't know for what purpose we have added that. But, I don't see any role here.
assert vhosts[-2] == '...done.'
This assert checks whether it has 'done...' in the output. But, for this version, this done is not even present in the output

thavanathan (approved):

saurabhkhanvilkar (approved):

low-level/framework/sspl_rabbitmq_reinit
```diff
@@ -257,7 +257,7 @@ def _create_vhost_if_necessary(virtual_host, product):
             response = _send_command(command)
 
         vhosts = response.split('\n')
-        assert vhosts[0] == 'Listing vhosts ...'
+        # assert vhosts[0] == 'Listing vhosts ...'
```
ujjwalpl, line 260: We should convert these into log prints, i.e. print vhosts or other values if not done already. In case config fails we will have this data for inspection.

ujjwalpl (changes requested):

low-level/framework/sspl_rabbitmq_reinit
Madhura-08: Done.

low-level/framework/sspl_rabbitmq_reinit
```diff
@@ -257,8 +257,8 @@ def _create_vhost_if_necessary(virtual_host, product):
             response = _send_command(command)
 
         vhosts = response.split('\n')
-        assert vhosts[0] == 'Listing vhosts ...'
         for vhost in vhosts[1:-1]:
+            print(vhosts)
```
cdeshmukh, line 261: this would print same 'vhosts' in loop, i think you meant 'vhost'. Or just print vhosts before loop, better

low-level/files/opt/seagate/sspl/bin/sspl_config
```diff
@@ -185,15 +185,28 @@ case $cmd in
         ;;
 esac
 
+rabbitmq_version=$(rpm -qi rabbitmq-server | awk -F': ' '/Version/ {print $2}')
```
cdeshmukh, line 188: > major_release . minor_release . maintenance_release

use this nomenclature below so its relatable easily

low-level/files/opt/seagate/sspl/bin/sspl_config
```diff
@@ -185,15 +185,28 @@ case $cmd in
...
 # Skip this step if sspl is being configured for node replacement scenario as consul data is already
 # available on healthy node
 # Updating RabbitMQ cluster nodes.
 # In node replacement scenario, avoiding feeding again to avoid over writing already configured values
 # with which rabbitmq cluster may have been created
 [ -f $REPLACEMENT_NODE_ENV_VAR_FILE ] || {
     # Update cluster_nodes key in consul
-    out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
-    pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
+    if [ $rabbitmq_base_version -eq 3 ] && [ $rabbitmq_middle_version -ge 8 ];
```
cdeshmukh, line 200: [ $rabbitmq_middle_version -eq 8 ]

would be  better, so we dont know if minor version 9 would support same format or not. if we upgrade

cdeshmukh (changes requested):

low-level/framework/sspl_rabbitmq_reinit
Madhura-08: That vhosts output will be difficult to understand if gets printed directly..S, printing it within a loop only but will be displayed as:
vhost: <Name of vhost>

low-level/files/opt/seagate/sspl/bin/sspl_config
Madhura-08: Ok

low-level/files/opt/seagate/sspl/bin/sspl_config
Madhura-08: Done

low-level/files/opt/seagate/sspl/bin/sspl_config
```diff
@@ -185,15 +185,35 @@ case $cmd in
...
 [ -f $REPLACEMENT_NODE_ENV_VAR_FILE ] || {
     # Update cluster_nodes key in consul
-    out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
-    pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
+    if [ $rabbitmq_major_release -eq 3 ] && [ $rabbitmq_maintenance_release -eq 8 ];
+    then
+        rabbitmq_cluster_status=$(/usr/sbin/rabbitmqctl cluster_status --formatter json)
+        cluster_running_nodes=$(python3 -c "rabbitmq_status=$rabbitmq_cluster_status; print(rabbitmq_status['running_nodes'])")
+        pout=$(echo $cluster_running_nodes | sed 's/rabbit@//g' | cut -d '[' -f2 | cut -d ']' -f1 | tr -d \'\" | tr -d ' ')
+    else
```
cdeshmukh, line 212: this may get applied to any version now, above 3.8.9 or below, and we intend to keep this for backward compatibility with 3.3.5 only, so better put that explicit check or fail out early with proper clear error string.

indrajitzagade: Can we avoid this parsing based on the version? Instead we can get what we back in the cluster status output: we could get:
running_nodes OR
Running Nodes

Then we can parse accordingly?

low-level/framework/sspl_rabbitmq_reinit
cdeshmukh: same vhosts will be printed multiple times, better before loop then.

low-level/files/opt/seagate/sspl/bin/sspl_config
Madhura-08: In the previous patch, that is why i kept it -ge (greater that or equal to). And they are giving this option --formatter json which must be with purpose (to get the output in json format which is widely accepted). So, I don't think that they will remove this option from next version.
With your suggested check, It will not work then if we go for other some check and eventually we will add more if else statements...

low-level/framework/sspl_rabbitmq_reinit
Madhura-08: I think I changed the approach now in the latest patch.

low-level/framework/sspl_rabbitmq_reinit
```diff
@@ -257,8 +257,8 @@ def _create_vhost_if_necessary(virtual_host, product):
             response = _send_command(command)
 
         vhosts = response.split('\n')
-        assert vhosts[0] == 'Listing vhosts ...'
         for vhost in vhosts[1:-1]:
+            print(f'vhost: {vhost}')
```
cdeshmukh, line 261: printing vhosts before loop to all values in one go. loop is breaking on one condition, in which case wont see all configured values

low-level/files/opt/seagate/sspl/bin/sspl_config
Madhura-08: @indrajitzagade Earlier version it was running_version, Now it is Running Nodes, tomorrow it can be else. 
From 3.8.9, they are giving this --formatter option in order to get the cluster status format in a united manner and json is widely used and easy for parsing. So, I think we should use that one.

low-level/framework/sspl_rabbitmq_reinit
```diff
@@ -324,10 +322,10 @@ def _create_user_if_necessary(username, password, virtual_host, product):
         response = _send_command(command)
 
         users = response.split("\n")
-        assert users[0] == 'Listing users ...'
         found_user = False
         for userspec in users[1:-1]:
             user = userspec.split()[0]
+            print(f'User: {user}')
```
cdeshmukh, line 328: printing 'users' before loop to all values in one go. loop is breaking on one condition, in which case wont see all configured values

low-level/framework/sspl_rabbitmq_reinit
```diff
@@ -343,12 +341,10 @@ def _create_user_if_necessary(username, password, virtual_host, product):
         users = subprocess.check_output(
             [RABBITMQCTL, 'list_users']
             ).decode("utf-8").split('\n')
-        assert users[0] == 'Listing users ...'
-        assert users[-2] == '...done.'
-        assert users[-1] == ''
         found_user = False
         for userspec in users[1:-1]:
             user = userspec.split()[0]
+            print('User: {user}')
```
cdeshmukh, line 347: same comment as line 328

low-level/files/opt/seagate/sspl/bin/sspl_config
indrajitzagade: oh ok @Madhura-08 ,this --formatter option looks useful here.

low-level/framework/sspl_rabbitmq_reinit
Madhura-08: Done

low-level/framework/sspl_rabbitmq_reinit
Madhura-08: Done

low-level/files/opt/seagate/sspl/bin/sspl_config
cdeshmukh: > From 3.8.9, they are giving this --formatter option in order to get the cluster status format in a united manner

--formatter with json choice, how it looks. And since they are changing with different versions, its not sure they would not change in future too, so having 'ge' may or may not work and we wont even know from logs what went wrong, which am trying to address here, by having version specific checks and beyond that give warning log, may be try once extracting running nodes and if failed, then come out OR come out directly without trying.

Also for 3.3.5 older version, this option is not there right, so we need to have that version based parsing still i feel.

Madhura-08: >Also for 3.3.5 older version, this option is not there right, so we need to have that version based parsing still i feel.
Right.
So, my check is:
if Version is >= 3.8.9
   use --formatter
else:
    use old method
So, 3.3.5 check will be satisfied here. 
And since they are changing with different versions
I think, its something additional they have added with some purpose. So, I don't think so, they will remove that thing in the next version.
And, suppose i added that as per your suggestion, it will be like below:
if Version is == 3.8.9
   use --formatter
else:if version == 3.3.5 (or may be <=):
    use old method
else:
    ??
What will be in else ??

cdeshmukh: >if Version is == 3.8.9
use --formatter

Agree

>else:if version == 3.3.5 (**dont add <= where we dont know the behavior really**):
use old method
else:
??
What will be in else ??

failout with an error log string

Madhura-08: Ok. I will do as you said then.

low-level/files/opt/seagate/sspl/bin/sspl_config
```diff
@@ -185,15 +185,39 @@ case $cmd in
...
+    then
+        rabbitmq_cluster_status=$(/usr/sbin/rabbitmqctl cluster_status --formatter json)
+        cluster_running_nodes=$(python3 -c "rabbitmq_status=$rabbitmq_cluster_status; print(rabbitmq_status['running_nodes'])")
+        pout=$(echo $cluster_running_nodes | sed 's/rabbit@//g' | cut -d '[' -f2 | cut -d ']' -f1 | tr -d \'\" | tr -d ' ')
+    elif [ $rabbitmq_version == "3.3.5" ];
+    then
+        out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
+        pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
+    else
+        echo "Not a supported version"
```
cdeshmukh, line 217: Not supported rabbitmq version %s (print version)

Madhura-08: Was doing the same. Done

cdeshmukh (approved):

indrajitzagade (approved):

low-level/files/opt/seagate/sspl/bin/sspl_config
```diff
@@ -185,15 +185,39 @@ case $cmd in
...
 # In node replacement scenario, avoiding feeding again to avoid over writing already configured values
 # with which rabbitmq cluster may have been created
 [ -f $REPLACEMENT_NODE_ENV_VAR_FILE ] || {
     # Update cluster_nodes key in consul
-    out=$(rabbitmqctl cluster_status | grep running_nodes | cut -d '[' -f2 | cut -d ']' -f1 | sed 's/rabbit@//g' | sed 's/,/, /g')
-    pout=$(echo $out | sed  "s/'//g" | sed  "s/ //g")
+    if [ $rabbitmq_major_release -eq 3 ] && [ $rabbitmq_maintenance_release -eq 8 ];
+    then
+        rabbitmq_cluster_status=$(/usr/sbin/rabbitmqctl cluster_status --formatter json)
+        cluster_running_nodes=$(python3 -c "rabbitmq_status=$rabbitmq_cluster_status; print(rabbitmq_status['running_nodes'])")
```
ujjwalpl, line 210: Please move all the rabbitmq related code into a function. It can be called as a plugin and the logic will work for kafka as well.

ujjwalpl (approved):

low-level/files/opt/seagate/sspl/bin/sspl_config
Madhura-08: Whatever I understood, I have done that. Please check now

ujjwalpl (approved):

PR #240, opened 2020-11-04
Store minion id, consul host and port in conf file during sspl_config

mohitpathak1:
## Problem Statement
<pre>
  <code>
    Story Ref (if any):
    EOS-14233
  </code>
</pre>
## Problem Description
<pre>
  <code>
    After node replacement, sspl is in stopped state on replaced node
  </code>
</pre>
## Solution
<pre>
  <code>
    Store minion id, consul host and port in conf file during sspl_config
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
    - [ ] No
  </code>
</pre>
<pre>
  <code>
  </code>
</pre>

### DIFF
--- a/low-level/framework/base/sspl_constants.py
+++ b/low-level/framework/base/sspl_constants.py
@@ -61,10 +61,12 @@ salt_uniq_passwd_per_node = ['RABBITMQINGRESSPROCESSOR', 'RABBITMQEGRESSPROCESSO
 
 # Initialize to default values
 node_key_id = 'srvnode-1'
-CONSUL_HOST = 'consulhost'
+CONSUL_HOST = '127.0.0.1'
 CONSUL_PORT = '8500'
 CONSUL_ERR_STRING = '500 No cluster leader'
 
+# TODO Keep only constants in this file.
+# other values(configs) should come from cofig.
 # Check if SSPL is configured
 if os.path.exists(SSPL_CONFIGURED):
     try:
--- a/low-level/framework/base/sspl_constants.py
+++ b/low-level/framework/base/sspl_constants.py
@@ -15,12 +15,14 @@
 # about this software or licensing, please email opensource@seagate.com or
 # cortx-questions@seagate.com.
 
+import configparser
 from enum import Enum
+import os
 
 try:
-    from salt_util import node_id, consulhost, consulport
+    from salt_util import SaltInterface
 except Exception as e:
-    from framework.utils.salt_util import node_id, consulhost, consulport
+    from framework.utils.salt_util import SaltInterface
 
 PRODUCT_NAME = 'LDR_R1'
 PRODUCT_FAMILY = 'cortx'
@@ -31,6 +33,7 @@ setups = ["cortx"]
 RESOURCE_PATH = f"/opt/seagate/{PRODUCT_FAMILY}/sspl/resources/"
 CLI_RESOURCE_PATH = f"/opt/seagate/{PRODUCT_FAMILY}/sspl/cli"
 DATA_PATH = f"/var/{PRODUCT_FAMILY}/sspl/data/"
+SSPL_CONFIGURED=f"/var/{PRODUCT_FAMILY}/sspl/sspl-configured"
 NODE_ID = "001"
 SITE_ID = "001"
 RACK_ID = "001"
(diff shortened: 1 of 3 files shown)


low-level/framework/base/sspl_constants.py
```diff
@@ -49,10 +52,39 @@
...
+# required only for init
+component = 'sspl/config'
+file_store_config_path = '/etc/sspl.conf'
+salt_provisioner_pillar_sls = 'sspl'
+salt_uniq_attr_per_node = ['cluster_id']
+salt_uniq_passwd_per_node = ['RABBITMQINGRESSPROCESSOR', 'RABBITMQEGRESSPROCESSOR', 'LOGGINGPROCESSOR']
+
+# Initialize to default values
+node_key_id = 'srvnode-1'
+CONSUL_HOST = 'consulhost'
```
indrajitzagade, line 64: is this default 127.0.0.1?

low-level/framework/utils/salt_util.py
```diff
@@ -14,83 +14,51 @@
 # cortx-questions@seagate.com.
 
 
-'''Module which provides common config information using Provisioner API'''
-
-
-import errno
-
-try:
-    from service_logging import logger
```
indrajitzagade, line 23: Why we moved from logger to print?

low-level/framework/utils/salt_util.py
```diff
@@ -100,28 +68,49 @@ def _is_server_single(self, _err=None):
...
+            if self.pillar_storage.get(f'{self.CLUSTER_KEY}:{self.CLUSTER_TYPE_KEY}') == 'single':
+                _is_single_node = True
         except Exception as err:
-            logger.warning(f'salt_util, Fail to read cluster type with \
-                             this error: {err}, Assuming single node')
+            print(f'salt_util, Fail to read cluster type with' + \
+                             f'this error: {err}, Assuming multi node')
         return _is_single_node
 
+    def is_env_vm(self):
```
indrajitzagade, line 78: This and below method is already there in utility.py file, why are we duplicating it here?

low-level/framework/utils/salt_util.py
```diff
@@ -100,28 +68,49 @@ def _is_server_single(self, _err=None):
...
+            if result == [] or result == "":
+                print("Not able to read whether env is vm or not, assuming VM env.")
+            else:
+                if 'false' in result[0].decode():
+                    is_vm = False
+        except Exception as e:
+            print(f"Error while reading whether env is vm or not, assuming VM env : {e}")
+        return is_vm
+
+    def execute_cmd(self, command=[]):
```
indrajitzagade, line 96: Same comment as line #78 above.

low-level/framework/utils/salt_util.py
mohitpathak1: When we call salt_util.py from sspl_config, it is called as stand alone python script... so the goal was to not import anything else from cortx-monitor otherwise the import will fail

low-level/framework/utils/salt_util.py
mohitpathak1: To avoid import failures when salt_util is called as a standalone python script from sspl_config

low-level/framework/base/sspl_constants.py
sandeepsgt: It should be 127.0.0.1. Done

low-level/framework/utils/salt_util.py
indrajitzagade: ok

low-level/framework/utils/salt_util.py
indrajitzagade: ok

indrajitzagade (approved):

saurabhkhanvilkar (approved):

low-level/framework/base/sspl_constants.py
```diff
@@ -49,10 +52,39 @@
...
+salt_uniq_passwd_per_node = ['RABBITMQINGRESSPROCESSOR', 'RABBITMQEGRESSPROCESSOR', 'LOGGINGPROCESSOR']
+
+# Initialize to default values
+node_key_id = 'srvnode-1'
+CONSUL_HOST = '127.0.0.1'
+CONSUL_PORT = '8500'
 CONSUL_ERR_STRING = '500 No cluster leader'
+
+# Check if SSPL is configured
+if os.path.exists(SSPL_CONFIGURED):
```
cdeshmukh, line 69: ideally we dont to maintain too much logic in this constants file, as its meant for constants specifically. See if we can maintain in different file & then just import final values. 
Not a must right now, keep a TODO if needed

low-level/framework/utils/salt_util.py
```diff
@@ -14,83 +14,51 @@
...
-                logger.error(
-                    f'Problem occured while reading from salt file. \
-                     File path: {SaltInterface.SALT_FILE} does not exist')
-            elif err == errno.EACCES:
-                logger.error(
-                    f'Problem occured while reading from salt file. \
-                     Permission issue while reading from: {SaltInterface.SALT_FILE}')
+            CMD = ['sudo', 'salt-call', '--local', 'grains.get', 'id', '--output=newline_values_only']
+            _result, _err, _ret_code = self.execute_cmd(CMD)
+            if _ret_code == 0 and _err.decode('utf-8') == '':
```
cdeshmukh, line 54: we need to start reading 'error' only if return code is non-zero and then check specific things in error string if needed. untangle these checks and depend on return code mainly

cdeshmukh (changes requested):
Need minor changes as given in review comments

low-level/framework/base/sspl_constants.py
sandeepsgt: @cdeshmukh `its meant for constants specifically` shouldn't those values which are not constant(values which are coming from some file based on logic) come from config ?

low-level/framework/utils/salt_util.py
sandeepsgt: Done

cdeshmukh (approved):

PR #344, opened 2021-03-16
EOS-18545: Move the common code out of sensors/message handlers into the Utils.

gargpalak048:
## Problem Statement
<pre>
  <code>
    Story Ref (if any):
    Please add Problem statement here...
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Please add Problem decription here...
  </code>
</pre>
## Solution
<pre>
  <code>
    Please add summary of the change,List any dependencies that are required for this change.
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
    - [ ] No
  </code>
</pre>
<pre>
  <code>
  </code>
</pre>

### DIFF
--- a/low-level/actuators/impl/generic/node_hw.py
+++ b/low-level/actuators/impl/generic/node_hw.py
@@ -27,7 +27,7 @@ from actuators.impl.actuator import Actuator
 from framework.base.debug import Debug
 from framework.utils.service_logging import logger
 from framework.base.sspl_constants import AlertTypes, SensorTypes, SeverityTypes, COMMON_CONFIGS
-from framework.utils.conf_utils import Conf, CLUSTER, GLOBAL_CONF, SITE_ID, RACK_ID, NODE_ID, SRVNODE
+from framework.base.global_config import GlobalConf
 
 
 class NodeHWactuator(Actuator, Debug):
@@ -36,9 +36,6 @@ class NodeHWactuator(Actuator, Debug):
 
     ACTUATOR_NAME = "NodeHWactuator"
     SYSTEM_INFORMATION = "SYSTEM_INFORMATION"
-    SITE_ID = "site_id"
-    RACK_ID = "rack_id"
-    NODE_ID = "node_id"
     NODE_REQUEST_MAP = {
         "disk" : "Drive Slot / Bay",
         "fan" : "Fan",
@@ -52,15 +49,13 @@ class NodeHWactuator(Actuator, Debug):
 
     def __init__(self, executor, conf_reader):
         super(NodeHWactuator, self).__init__()
-        self._site_id = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{SITE_ID}",'DC01')
-        self._rack_id = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{RACK_ID}",'RC01')
-        self._node_id = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{NODE_ID}",'SN01')
         self.host_id = socket.getfqdn()
         self.sensor_id_map = None
         self._executor = executor
         self.fru_specific_info = {}
         self._resource_id = ""
         self._sensor_type = ""
+        self._conf = GlobalConf().fetch_global_config(['site_id', 'node_id', 'rack_id'])
 
     def initialize(self):
         """Performs basic Node HW actuator initialization"""
@@ -177,9 +172,9 @@ class NodeHWactuator(Actuator, Debug):
           "host_id": self.host_id,
           "instance_id": resource_id,
           "info": {
-            "site_id": self._site_id,
-            "rack_id": self._rack_id,
-            "node_id": self._node_id,
+            "site_id": self._conf['site_id'],
+            "rack_id": self._conf['rack_id'],
+            "node_id": self._conf['node_id'],
             "resource_id": resource_id,
             "resource_type": resource_type,
             "event_time": epoch_time
@@ -275,9 +270,9 @@ class NodeHWactuator(Actuator, Debug):
         response['alert_type'] = AlertTypes.GET.value
         response['severity'] = SeverityTypes.INFORMATIONAL.value
         response['info'] = {
-            "site_id": self._site_id,
-            "rack_id": self._rack_id,
-            "node_id": self._node_id,
+            "site_id": self._conf['site_id'],
+            "rack_id": self._conf['rack_id'],
+            "node_id": self._conf['node_id'],
             "resource_type": "node:sensor:" + self._sensor_type.lower(),
             "resource_id": self._resource_id,
             "event_time": str(calendar.timegm(time.gmtime())),
--- a/low-level/files/opt/seagate/sspl/setup/generate_resource_health_view/resource_health_view
+++ b/low-level/files/opt/seagate/sspl/setup/generate_resource_health_view/resource_health_view
@@ -38,10 +38,7 @@ from framework.utils.ipmi_client import IpmiFactory
 from framework.utils.sysfs_interface import SysFS
 from framework.utils.tool_factory import ToolFactory
 from framework.base.sspl_constants import PRODUCT_FAMILY , DATA_PATH
-from framework.utils.conf_utils import (CLUSTER, ENCLOSURE, GLOBAL_CONF, SRVNODE,
-    STORAGE, Conf, NODE_TYPE, TYPE, SITE_ID, RACK_ID, NODE_ID, CLUSTER_ID, RELEASE,
-    TARGET_BUILD)
-
+from framework.base.global_config import GlobalConf
 
 class SSPLHealthView(object):
     """docstring for SSPLHealthView"""
@@ -62,7 +59,11 @@ class SSPLHealthView(object):
     SERVER_TYPE_PHYSICAL = 'physical'
 
     def __init__(self):
-
+        self._global_conf = GlobalConf()
+        self._conf = self._global_conf.fetch_global_config(
+            ['cluster_id','site_id', 'node_id', 'rack_id', 'target_build'])
+        self._cortx_build = self._conf['target_build']
+        self._srvnode = self._global_conf.SRVNODE
         persistent_data_location = DATA_PATH
         storage_encl_dir = persistent_data_location + 'encl/frus/'
         self.encl_disks_filename = storage_encl_dir + 'disks/'
@@ -96,11 +97,7 @@ class SSPLHealthView(object):
 
         self.encl_hw_specifics_api = 'configuration'
 
-        self.SITE_ID = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{SITE_ID}", 'DC01')
-        self.RACK_ID = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{RACK_ID}", 'RC01')
         self.HOST_NAME = socket.getfqdn()
-        self.NODE_ID = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{NODE_ID}", 'SN01')
-        self.CLUSTER_ID = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{CLUSTER_ID}", 'CC01')
         self.fru_id_map = None
         self.sensor_id_map = None
         self.ipmi_fru_lst = ['disk', 'fan', 'psu']
@@ -125,13 +122,13 @@ class SSPLHealthView(object):
                 "version": "2.0"
             }
         })
-        self._resource_health_view['cluster'].update({'cluster_id':self.CLUSTER_ID})
-        self._resource_health_view['cluster']['sites'][self.SITE_ID] =\
+        self._resource_health_view['cluster'].update({'cluster_id':self._conf['cluster_id']})
+        self._resource_health_view['cluster']['sites'][self._conf['site_id']] =\
             self._resource_health_view['cluster']['sites'].pop('site_id')
 
-        self._resource_health_view['cluster']['sites'][self.SITE_ID]['rack']\
-            [self.RACK_ID] = self._resource_health_view['cluster']['sites']\
-                [self.SITE_ID]['rack'].pop('rack_id')
+        self._resource_health_view['cluster']['sites'][self._conf['site_id']]['rack']\
+            [self._conf['rack_id']] = self._resource_health_view['cluster']['sites']\
+                [self._conf['site_id']]['rack'].pop('rack_id')
 
     def run(self, args):
         health_view_created = False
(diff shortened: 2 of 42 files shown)


cortx-admin:

low-level/framework/utils/conf_utils.py
```diff
@@ -148,3 +152,8 @@
 
 SRVNODE = Conf.get(GLOBAL_CONF, f'{CLUSTER}>{SERVER_NODES}')[MACHINE_ID]
 ENCLOSURE = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{STORAGE}>{ENCLOSURE_ID}")
+
+SITE_ID = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{'site_id'}",'DC01')
```
cdeshmukh, line 156: define default values for all these four in sspl_constants.py and use from there.
DataCenter01
Rack01
ServerNode01
Cluster01

cdeshmukh, line 156: Keeping them on conf_utils doesnt makes sense with these changes, as these 4 entries no more relates to any util code. We can create a new file named global_config.py under sspl/low-level/framework/base, where all these cortx global config can be maintained / updated and made available for rest of the code, like sensors & actuators.

low-level/framework/utils/utility.py
```diff
@@ -80,6 +83,54 @@ def get_os(self):
...
+
+    def read_raw_filedata(self, file):
+        """Reads the data from the file and returns it as a string."""
+        if not os.path.isfile(file):
+            return str(None)
+
+        with open (file, "r") as datafile:
+            return str(datafile.read().replace('\n', ''))
+
+    def get_hostname(self):
```
cdeshmukh, line 108: hope these functions added here are nothing but copy of duplicated but tested code historically. we would like to use that code obviously, to not to invent the wheel again

low-level/framework/utils/utility.py
```diff
@@ -80,6 +83,54 @@ def get_os(self):
             else:
                 return None
 
+    def get_epoch_time(self, date, _time):
```
cdeshmukh, line 86: Also address these codacy warnings, for new code atleast we should be targeting
https://docs.quantifiedcode.com/python-anti-patterns/correctness/method_could_be_a_function.html

low-level/framework/utils/conf_utils.py
gargpalak048: So, In the global_config.py we will load the global_conf file. (#L149-151) and all the entites like MACHINE_ID, ENCLOSURE_ID, STORAGE etc will get imported from conf_utils.py.
Please confirm

low-level/framework/utils/utility.py
gargpalak048: Yes.. Currently these functions are the duplicated ones.. Later we can remove the functions from seperate files and directly can use it from here.

low-level/framework/utils/conf_utils.py
gargpalak048: Done.

low-level/framework/utils/utility.py
gargpalak048: 'Method could be a function' comes for the ambiguous functions, which does not refer to any instance or class.
Here these functions are using self keyword.
Not sure why still this warning coming.
Except this warning, addressed other codacy warnings

low-level/framework/utils/utility.py
cdeshmukh: Yes ok.

low-level/framework/utils/conf_utils.py
cdeshmukh: Yes

low-level/actuators/impl/generic/node_hw.py
```diff
@@ -27,7 +27,7 @@
 from framework.base.debug import Debug
 from framework.utils.service_logging import logger
 from framework.base.sspl_constants import AlertTypes, SensorTypes, SeverityTypes, COMMON_CONFIGS
-from framework.utils.conf_utils import Conf, CLUSTER, GLOBAL_CONF, SITE_ID, RACK_ID, NODE_ID, SRVNODE
+from framework.base.global_config import CLUSTER_ID, SITE_ID, RACK_ID, NODE_ID
```
cdeshmukh, line 30: Avoid CLUSTER_ID if not used, address all such codacy issues

low-level/framework/base/global_config.py
```diff
@@ -0,0 +1,36 @@
...
+            ENCLOSURE_ID)
+from framework.base.sspl_constants import (DATACENTER01,
+                                           RACK01, SERVERNODE01,
+                                           CLUSTER01)
+
+Conf.load(SSPL_CONF, "yaml:///etc/sspl.conf")
+global_config = Conf.get(SSPL_CONF, "SYSTEM_INFORMATION>global_config_copy_url")
+Conf.load(GLOBAL_CONF, global_config)
+
+SRVNODE = Conf.get(GLOBAL_CONF, f'{CLUSTER}>{SERVER_NODES}')[MACHINE_ID]
```
cdeshmukh, line 30: One discussion today with Ujjwal altered the whole meaning of having this PR. We want to **avoid duplicated code**, so we pulling common code in one file. However, config values here pulled once using confstor apis and then made available throughout life of the sspl service.

**Problem is what if value changes dynamically ?** Confstor apis would reflect changed values every time we call it, while here with this approach it wont. 

Now how do we address both ? Again a wrapper functions in this file to call confstor api kv pair we need, instead better is existing code then, which directly calls confstor api per sensor.

I dont know how to tackle this yet, better we think through & then implement. Originally we were limiting to new sensor code only, then were planning to extend later for existing sensors (since we would know the implications for new approach too)

cdeshmukh: how about one function in here to fetch common type of configs using confstore apis, like

cluster_id, site_id, rack_id, machine_id= get_resource_location(fetch_cluster_id=false, fetch_site_id=false, fetch_rack_id=false, fetch_machine_id=false)

With filters to fetch any required id even, like
clusterid, siteid, , = get_resource_location(true,true, false, false)

Same way other functions we can have more functions in future,  to fetch more global configs in group together, like for storage_set, servers_in_storageset, enclosures_in_storage_set

These functions call per sensor would be needed to fetch required configs then !

**This is just one proposal, feel to add better ones**

gargpalak048: To generalize the code, I am thinking we can keep two dict.
one global dict will map the key with val as a tuple of req fields.
For eg:-
        config_dict = {
            "cluster_id": (CLUSTER, SRVNODE, 'site_id')
        }
And one more dict to have default values of the key
for eg:- 
        default_val = {
            "cluster_id": 'CC01'
        }		


And a function which will take a config_list required by diff sensors
and will process the request and will return the response dict.
	def fetch_config_from_global_dict(self, config_list):
        response = {}
        default = ""
        for key in config_list:
            if key not in config_dict.keys():
                return
            val = config_dict[key]
            if key in default_val.keys():
                default = default_val[key]
            req_str =  '>'.join(val)
            global_val = Conf.get(GLOBAL_CONF, req_str, default)
            response[key] = global_val
        return response

But with this approach we have to maintain 2 dicts and have to map the req fields by sensors to this dict.
@cdeshmukh @sandeepsgt  please review this approach, if it seems too complex, we can proceed with chetan's approach.

cdeshmukh: Am not totally clear on this approach, we can discuss.

But definite take away is to keep the list of keys which can be requested, as flexible.

And we also want to take away the complexity & repeated code of requesting common/sspl config from consumer layers like sensors & actuators.

So we can two main functions
<return dict wih kv pairs> get_sspl_config(<set of keys as i/p>)
<return dict wih kv pairs> get_global_config(<set of keys as i/p>)

Now both these functions would take just keys (with hierarchies) and internally use confstor apis to fetch from respective backends (config index). So sensors/actuators will still have to import keys as its done today.

low-level/actuators/impl/generic/node_hw.py
```diff
@@ -367,3 +363,11 @@ def _response_to_dict(self, data, split_char=':', dict_keys=None, many_sensors=F
             return many_sensors_data
 
         return properties
+
+    def _read_config(self):
+        """ Read Configurations from global config"""
+        conf_list = ['site_id', 'node_id', 'rack_id']
+        response_dict = self._global_conf.fetch_global_config(conf_list)
```
cdeshmukh, line 370: this is infact increasing the code compared to previously.

Why cant we have like
loc  =  self._global_conf.fetch_global_config(['site_id', 'node_id', 'rack_id'])

and use loc['site_id'] directly where ever needed. Anyways it would have default have if error occurs while fetching an config.

low-level/files/opt/seagate/sspl/setup/generate_resource_health_view/resource_health_view
```diff
@@ -1232,8 +1239,10 @@ if __name__ == "__main__":
         print('Missing arguments')
         parser.print_help()
         sys.exit(1)
-    storage_type = Conf.get(GLOBAL_CONF, f"{STORAGE}>{ENCLOSURE}>{TYPE}", 'virtual').lower()
-    server_type = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{SRVNODE}>{NODE_TYPE}", 'virtual').lower()
+    
+    response_dict = GlobalConf().fetch_global_config(["storage_type", "server_type"])
+    storage_type = response_dict['storage_type']
```
cdeshmukh, line 1244: Keep response_dict name shorter, like resp and refer elements directly as res['storage_type']

low-level/framework/base/global_config.py
```diff
@@ -0,0 +1,104 @@
...
+            USER, PRODUCT, RSYSLOG, HOST, NODE_TYPE, TYPE, RELEASE,
+            TARGET_BUILD, SECRET, CONF_API_DELIMITER)
+from framework.base.sspl_constants import (DATACENTER01, RACK01, SERVERNODE01,
+                CLUSTER01, DEFAULT_MC_IP, DEFAULT_SYSLOG_HOST, DEFAULT_SYSLOG_PORT)
+
+
+class GlobalConf:
+    """Base Class to fetch global config values."""
+
+    Conf.load(SSPL_CONF, "yaml:///etc/sspl.conf")
```
cdeshmukh, line 31: why this code(line 31-33) here? can stay in __init__

low-level/framework/base/global_config.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self):
+        self.initialize()
+        self.map_global_config_key_values()
+
+    def initialize(self):
+        self.SRVNODE = Conf.get(GLOBAL_CONF, f'{CLUSTER}>{SERVER_NODES}')[MACHINE_ID]
+        self.ENCLOSURE = Conf.get(GLOBAL_CONF, f"{CLUSTER}>{self.SRVNODE}>{STORAGE}>{ENCLOSURE_ID}")
+
+    def fetch_global_config(self, config_list):
+        response = {}
```
cdeshmukh, line 44: add function description & keep a blank line after

low-level/framework/base/global_config.py
```diff
@@ -0,0 +1,104 @@
...
+        default_val = ''
+        req_fields = []
+        for key in config_list:
+            if key not in self.global_conf_dict.keys():
+                return
```
cdeshmukh, line 49: log this with logger.debug

low-level/framework/base/global_config.py
```diff
@@ -0,0 +1,104 @@
...
+        return response
+
+    def fetch_sspl_config(self, query_string=False, default_val=""):
+        try:
+            if not query_string:
+                return
+            response = Conf.get(SSPL_CONF, query_string, default_val)
+            return response
+        except Exception as ex:
+            print("Failed to fetch the sspl config:{}".format(str(ex)))
```
cdeshmukh, line 69: Avoid print() unless independent scripts!! use logger() with proper log level, logger.error() here

low-level/framework/base/global_config.py
```diff
@@ -0,0 +1,104 @@
...
+                return
+            response = Conf.set(SSPL_CONF, request, value)
+            return response
+        except Exception as ex:
+            print("Failed to set the sspl config:{}".format(str(ex)))
+
+    def map_global_config_key_values(self):
+        self.global_conf_dict = {
+            "cluster_id": [CLUSTER, 'cluster_id', 'default:' + CLUSTER01],
+            "site_id": [CLUSTER, self.SRVNODE, 'site_id', 'default:' + DATACENTER01],
```
cdeshmukh, line 83: Good to see you found a way to keep default values in same table.

so "CLUSTER, self.SRVNODE" holds the key hierarchy and 'site_id' is the actual key right. site_id is the key to this list also, why duplicated in list?
It can stay as 
"site_id": [ CLUSTER + self.SRVNODE, 'default:' + DATACENTER01],

low-level/framework/base/sspl_constants.py
```diff
@@ -70,6 +70,13 @@
 SSPL_CONFIG_INDEX = "sspl"
 SSPL_TEST_CONFIG_INDEX = "sspl_test"
 CONFIG_SPEC_TYPE = "yaml"
+DATACENTER01 = "DC01"
```
cdeshmukh, line 73: name as DEFAULT_DC and alike

low-level/framework/platforms/realstor/realstor_enclosure.py
```diff
@@ -27,11 +27,9 @@
...
 from framework.target.enclosure import StorageEnclosure
 from framework.utils import encryptor
-from framework.utils.conf_utils import (CLUSTER, CONTROLLER, ENCLOSURE,
-                                        GLOBAL_CONF, IP, MGMT_INTERFACE,
-                                        SECRET, POLLING_FREQUENCY, PORT,
-                                        PRIMARY, SECONDARY, SRVNODE, SSPL_CONF,
-                                        STORAGE, STORAGE_ENCLOSURE, USER, Conf)
+from framework.base.global_config import GlobalConf
+from framework.utils.conf_utils import (MGMT_INTERFACE, POLLING_FREQUENCY,
+                                        STORAGE, STORAGE_ENCLOSURE)
```
cdeshmukh, line 32: why importing something which is not getting used? like raised by CODACY here?

low-level/actuators/impl/generic/node_hw.py
```diff
@@ -27,7 +27,7 @@
 from framework.base.debug import Debug
 from framework.utils.service_logging import logger
 from framework.base.sspl_constants import AlertTypes, SensorTypes, SeverityTypes, COMMON_CONFIGS
-from framework.utils.conf_utils import Conf, CLUSTER, GLOBAL_CONF, SITE_ID, RACK_ID, NODE_ID, SRVNODE
+from framework.base.global_config import GlobalConf
```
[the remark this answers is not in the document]
gargpalak048: What about SSPL config, which ever sensor needs it, would se same way ?
I have two seperate functions in global_config.py.
fetch_global_config()
fetch_sspl_config()
As for global config the fields will remain same. So I am maintaining a global dict to map key & values.
But for sspl config values, every sensor will have different query string. so every sensor is calling the fetch_sspl_config() with params: (query_string, default_val)
Please refer to: 
https://github.com/gargpalak048/cortx-monitor/blob/EOS-18545-new/low-level/framework/base/global_config.py

low-level/framework/base/global_config.py
gargpalak048: For each sspl service process, we can't load the Conf more than once.
If we keep this block of code in init, whenever each sensor will create an obj, 
 Conf.load(GLOBAL_CONF, global_config)
 will be called, which raises the error.

low-level/framework/base/global_config.py
gargpalak048: currently in fetch_global_config() function,
the requested str is created as:
req_string =  CONF_API_DELIMITER.join(req_fields)
irrespective of the fields.
If we want to keep the dict as:
"site_id": [ CLUSTER + self.SRVNODE, 'default:' + DATACENTER01],
then we need to add additional code to add 'site_id' key in the req_field list before 'default:value' element.
Please suggest, should I do this change?

low-level/framework/base/global_config.py
cdeshmukh: we discussed this later & latest code reflects what was needed

PR #359, opened 2021-04-06
EOS-19184 - SSPL - User permission issue on /opt/seagate/cortx/sspl

mariyappanp:


low-level/files/opt/seagate/sspl/setup/sspl_prepare.py
```diff
@@ -103,7 +103,7 @@ def validate(self):
         # Validate network interface availability
         for i_list in [mgmt_interfaces, data_private_interfaces, data_public_interfaces]:
             self.validate_nw_cable_connection(i_list)
-        self.validate_nw_interfaces()
+            self.validate_nw_interfaces(i_list)
```
mariyappanp, line 106: @cdeshmukh 
Instead of validating all the nw interfaces in the system, this change will only validate the interfaces provided in the input config files.

low-level/sspl-ll.spec
```diff
@@ -133,7 +133,7 @@ rm -f /etc/dbus-1/system.d/sspl-ll_dbus_policy.conf
 rm -f /usr/bin/resource_health_view /usr/bin/sspl_bundle_generate /usr/bin/manifest_support_bundle
 
 %files
-%defattr(-,sspl-ll,root,-)
+%defattr(-,root,root,-)
```
cdeshmukh, line 136: this would change file / directories permissions for use by root user alone. sspl-ll user, with which sspl_ll_d runs, will have issues writing to those dirs/files again ? How sanity passing with these failures ?

low-level/files/opt/seagate/sspl/setup/sspl_prepare.py
cdeshmukh: yes, got it

low-level/sspl-ll.spec
cdeshmukh: > Mariyappan - This is handled in post_install where we provide ownership to sspl-ll after sspl-ll user creation.

Understood, fine then.

cdeshmukh (approved):

sandeepsgt (approved):

satish-seagate (approved):

PR #404, opened 2021-05-24
EOS-20685-RemoveIpmiHardCodedCmd : Use of general IPMI cmd executor in IPMI tool.

sumedhak27:
Signed-off-by: Sumedh A. Kulkarni <sumedh.a.kulkarni@seagate.com>

## Problem Statement:
Generate Support Bundle has hardcoded ipmi command and Node Hw sensor implements the ipmi cmd executor instead of using the IPMITool

## Solution Design:
Resign of IPMITool: _run_ipmitool_subcommand() function so that support bundle or any other sspl module can execute the IPMI command with abstraction.

## Coding:

* [x] Did you follow coding standard and verified it with flake8? 
* [ ] Did your address all Codacy issues?
* [x] Did you self review the PR?

Modified the IPMITool class, specifically the _run_ipmitool_subcommand() method to address general request to execute a IPMI command, used the generic function in node_hw sensor, actuator and support bundle generator.

## Testing:

* [x] Unittest updated and all unittests are passing?
* [x] Sanity/Self tests are updated (if applicable)?
* [ ] Manual testing done covering happy path and non-happy path?

## Integration:

* [ ] If there is any interface change, did you communicate it to other components?
* [ ] If there is any interface change, other componenent gate gatekeepers are ready to accept the change?
* [ ] Did you complete deployment test if applicable?

## PR checklist:
* [x] Did you add Jira number to the PR?

PR #410, opened 2021-06-04
EOS-17462: SSPL: Adopt new IEM framework

gargpalak048:
SSPL:Adopt new IEM framework

## Solution Design:
Design changes to generate & send IEM alerts using new framework

## Coding:

* [x ] Did you follow coding standard and verified it with flake8? 
* [ x] Did your address all Codacy issues?

* [ x] Sanity/Self tests are updated (if applicable)?

verify test iem alert received in specific format

## Integration:

* [x ] If there is any interface change, did you communicate it to other components?
* [ x] If there is any interface change, other componenent gate gatekeepers are ready to accept the change?

## PR checklist:
* [ x] Did you add Jira number to the PR?

yes

* [ x] DCO and cla-signed?


low-level/framework/utils/iem.py
```diff
@@ -155,6 +159,9 @@ def check_exsisting_fault_iems(self):
             if prev_fault_iem_event:
                 self.fault_iems.append(event_name)
 
-def generate_iem(iem):
-    """Generate iem."""
-    syslog.syslog(iem)
+    def generate_iem(self, module, event_code, severity, description, specific_info):
+        """Generate iem and send it to a MessgaeBroker."""
+
+        EventMessage.send(module=module, event_id=event_code,
```
cdeshmukh, line 165: no exception handling? We need to have it, even if EventMessage is not raising explicitly any !

low-level/sspl_ll_d
```diff
@@ -715,6 +716,9 @@ def send_thread_controller_request(state):
 #     logger.info(f"Received SIGUSR1 signal to set log level to '{log_level}'")
 #     logger.setLevel(log_level)
 
+def initialize_iem():
+    EventMessage.init(component='sspl', source='S')
```
cdeshmukh, line 720: again, capture errors. log.debug in case of success, log.error in case of error encountered

low-level/framework/utils/iem.py
gargpalak048: Done

low-level/sspl_ll_d
gargpalak048: Done

cortx-admin:

unittests/test_iem.py
```diff
@@ -0,0 +1,111 @@
...
+        self.assertEqual(fault_alert["iem"]["info"]["severity"], "Informational")
+        self.assertEqual(fault_alert["iem"]["source"]["module"], "ipmitool")
+        self.assertEqual(fault_alert["iem"]["contents"]["event"], "0050010002")
+
+    def tearDown(self):
+        pass
+
+def run_command(command):
+    """Run the command and get the response and error returned"""
+    process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jjxsg, line 102: `shell=True` is platform dependent and prone to shell injection, please avoid using it.

unittests/test_iem.py
```diff
@@ -0,0 +1,111 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License along
+# with this program. If not, see <https://www.gnu.org/licenses/>. For any questions
+# about this software or licensing, please email opensource@seagate.com or
+# cortx-questions@seagate.com.
+
+import os
+import sys
+import unittest
+import subprocess
```
jjxsg, line 19: Use  `SimpleProcess` from `cortx.utils.process` instead of `subprocess`

unittests/test_iem.py
gargpalak048: @jjxsg run_command func is getting used in other files like( disk_monitor.py, raid.py, disk_msg_handler.py etc) 
in the same manner.
Also I tried executing cmd without shell=True param, but it throws error.
Do you have any workaround?

jjxsg: @gargpalak048 please refer to first comment in this PR https://github.com/Seagate/cortx-monitor/pull/404
You need to split the command properly using `shlex` to be able to execute command without starting a shell.

gargpalak048: Thanks for the ref @jjxsg.
So in cortx-utils SimpleProcess class, this func already implemented.
Just parse the cmd and it will return the response.
No need to shlex here.

unittests/test_iem.py
gargpalak048: Done. Please review now

unittests/test_iem.py
jjxsg: @gargpalak048 Sorry, I didn't get you. Which function you are talking about? And I can see `subprocess` with `shell=True`

gargpalak048: Using SimpleProcess.run() instead of run_command() func.
Will remove this func from test script

low-level/framework/utils/iem.py
```diff
@@ -155,6 +153,12 @@ def check_exsisting_fault_iems(self):
...
-    """Generate iem."""
-    syslog.syslog(iem)
+    @classmethod
+    def generate_iem(cls, module, event_code, severity, description):
+        """Generate iem and send it to a MessgaeBroker."""
+        try:
+            EventMessage.send(module=module, event_id=event_code,
+                              severity=severity, message_blob=description)
+        except EventMessageError as e:
+            logger.error("Failed to generate & send IEM alert."
```
cdeshmukh, line 163: >Failed to send IEM alert

THis much suffice

low-level/files/opt/seagate/sspl/setup/sspl_dev_deploy
```diff
@@ -403,8 +404,9 @@ class SSPLDevDeploy:
             "%s post_install --config yaml://%s" % (utils_setup_script, post_install_config),
             "%s post_install --config yaml://%s" % (sspl_setup_script, post_install_config),
             "%s prepare --config yaml://%s" % (sspl_setup_script, prepare_config),
-            "%s config --config yaml://%s" % (utils_setup_script, config_config),
+            "%s config --config yaml://%s" % (utils_setup_script, utils_config_tmpl),
```
cdeshmukh, line 407: why this name change here. Similar names exists at other lines, signifying the mini provisioner stages i.e <mini-provi-stage>_config.
Now config_config is not proper i agree, though try to improve on that and not change the rule followed here.
Like can use 'miniprovConfig_config'

low-level/files/opt/seagate/sspl/setup/sspl_dev_deploy
```diff
@@ -403,8 +404,9 @@ class SSPLDevDeploy:
...
             "%s config --config yaml://%s" % (sspl_setup_script, config_config),
+            "%s init" % (utils_setup_script),
```
cdeshmukh, line 409: why two times init needs to be called, thats not proper. Merge required config in either miniprov Config or Init stage

low-level/sspl_ll_d
```diff
@@ -715,6 +717,15 @@ def send_thread_controller_request(state):
 #     logger.info(f"Received SIGUSR1 signal to set log level to '{log_level}'")
 #     logger.setLevel(log_level)
 
+def initialize_iem():
+    try:
+        EventMessage.init(component='sspl', source='S')
+        logger.debug("IEM framework initiated successfully!!!")
```
cdeshmukh, line 723: let this be logger.info(), useful info to see during sspl bootup

low-level/sspl_ll_d
```diff
@@ -715,6 +717,15 @@ def send_thread_controller_request(state):
...
+    except EventMessageError as e:
+        logger.error("Failed to initialize IEM Framework."
+                     f"ERROR:{e}")
+        raise
```
cdeshmukh, line 727: This would make sspl exit with traceback, since no code is handling exception from this function ?

low-level/sspl_ll_d
```diff
@@ -777,6 +788,9 @@ if __name__ == "__main__":
     signal.signal(signal.SIGUSR2, refresh_iem_log_file)
     #DO NOT EDIT: Marker comment to dynamically add signal handler for SIGUSR1 to generate code coverage report
 
+    # Initialize IEM framework
```
cdeshmukh, line 791: Needed under try-exception block. And should SSPL exit if this fails, probably not, to continue allowing other imp monitoring. Log it as critical, in bold about IEM and continue

cdeshmukh (changes requested):

low-level/framework/utils/iem.py
gargpalak048: Done

low-level/files/opt/seagate/sspl/setup/sspl_dev_deploy
gargpalak048: @cdeshmukh it's not a name change, different conf file is parsed.
Earlier same config file was parsed to both utils_setup & sspl_setup script.
As per cortx-utils wiki, we require an input variable file to parsed in the config phase.
so I kept utils config input file inside the /tmp folder 

#L381
> utils_config_tmpl = "/tmp/utils.config.tmpl.%s" % (self.deploy_type)

low-level/files/opt/seagate/sspl/setup/sspl_dev_deploy
gargpalak048: first init is for utils_setup and next init phase is of sspl_setup script

low-level/sspl_ll_d
gargpalak048: done

low-level/sspl_ll_d
gargpalak048: done the change

low-level/sspl_ll_d
gargpalak048: Ack.
logging the error with critical label and sspl will continue

low-level/files/opt/seagate/sspl/setup/sspl_dev_deploy
gargpalak048: Below is the cortx-utils mini provisioning wifi ref:
https://github.com/Seagate/cortx-utils/wiki/%22cortx-py-utils%22-single-node-manual-provisioning

cdeshmukh: oh i misunderstood, got it now. thanks

cdeshmukh (approved):

low-level/framework/utils/iem.py
```diff
@@ -155,6 +153,12 @@ def check_exsisting_fault_iems(self):
             if prev_fault_iem_event:
                 self.fault_iems.append(event_name)
 
-def generate_iem(iem):
-    """Generate iem."""
-    syslog.syslog(iem)
+    @classmethod
+    def generate_iem(cls, module, event_code, severity, description):
```
thavanathan, line 157: why do we need a classmethod? we can keep this as a staticmethod?

low-level/framework/utils/iem.py
```diff
@@ -155,6 +153,12 @@ def check_exsisting_fault_iems(self):
             if prev_fault_iem_event:
                 self.fault_iems.append(event_name)
 
-def generate_iem(iem):
-    """Generate iem."""
-    syslog.syslog(iem)
```
thavanathan, line 160: You can remove the syslog import.

low-level/sspl_ll_d
```diff
@@ -715,6 +717,15 @@ def send_thread_controller_request(state):
 #     logger.info(f"Received SIGUSR1 signal to set log level to '{log_level}'")
 #     logger.setLevel(log_level)
 
+def initialize_iem():
+    try:
+        EventMessage.init(component='sspl', source='S')
```
thavanathan, line 722: What is source argument?

low-level/sspl_ll_d
```diff
@@ -777,6 +788,13 @@ if __name__ == "__main__":
     signal.signal(signal.SIGUSR2, refresh_iem_log_file)
     #DO NOT EDIT: Marker comment to dynamically add signal handler for SIGUSR1 to generate code coverage report
 
+    # Initialize IEM framework
+    try:
```
thavanathan, line 792: try except block is also there in `initialize_iem` function. Either use `try/except` block here or in `initialize_iem` function.

unittests/test_iem.py
```diff
@@ -0,0 +1,105 @@
...
+import os
+import sys
+import unittest
+
+from cortx.utils.iem_framework import EventMessage
+from cortx.utils.process import SimpleProcess
+
+PROJECT_ROOT = "/".join(os.path.abspath(__file__).split("/")
+                        [:-2]) + "/low-level"
+sys.path.append(PROJECT_ROOT)
```
thavanathan, line 25: We don't need this. This path is already imported in `unittests/conftest.py` which is config file for pytest.

unittests/test_iem.py
```diff
@@ -0,0 +1,105 @@
...
+        self.module = None
+        self.event_code = None
+        self.EVENT_CODE = {
+        "IPMITOOL_ERROR" : ["0050010001", "ipmitool"],
+        "IPMITOOL_AVAILABLE" : ["0050010002", "ipmitool"]
+        }
+        EventMessage.init(component='sspl', source='S')
+
+    def mock_ipmitool_fault(self):
+        cmd = 'yum remove ipmitool'
```
thavanathan, line 39: Looks like a distructive operation.

unittests/test_iem.py
```diff
@@ -0,0 +1,105 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License along
+# with this program. If not, see <https://www.gnu.org/licenses/>. For any questions
+# about this software or licensing, please email opensource@seagate.com or
+# cortx-questions@seagate.com.
+
+import os
```
thavanathan, line 16: Please move this file to unittests/framework/utils/

Also there is no use of unittesting `cortx.utils.iem_framework`. We need to test framework/utils/iem.py. So please update calls to generate IEM to use framework/utils/iem.py.

unittests/test_iem.py
```diff
@@ -0,0 +1,105 @@
...
+        self.assertIs(type(fault_alert), dict)
+        self.assertEqual(fault_alert["iem"]["info"]["severity"], "Informational")
+        self.assertEqual(fault_alert["iem"]["source"]["module"], "ipmitool")
+        self.assertEqual(fault_alert["iem"]["contents"]["event"], "0050010002")
+
+    def tearDown(self):
+        pass
+
+
+if __name__ == "__main__":
```
thavanathan, line 101: test is failing for me. Please refer https://seagate-systems.atlassian.net/wiki/spaces/sspl/pages/273449515/Unit+Testing+Guide

thavanathan (changes requested):
Please fill PR template.

unittests/test_iem.py
```diff
@@ -0,0 +1,105 @@
...
+        self.event_code = None
+        self.EVENT_CODE = {
+        "IPMITOOL_ERROR" : ["0050010001", "ipmitool"],
+        "IPMITOOL_AVAILABLE" : ["0050010002", "ipmitool"]
+        }
+        EventMessage.init(component='sspl', source='S')
+
+    def mock_ipmitool_fault(self):
+        cmd = 'yum remove ipmitool'
+        res, _, _ = SimpleProcess(cmd).run()
```
jjxsg, line 40: You can directly call `SimpleProcess(cmd).run()` without capturing returned values if you don't have any use of those values.

low-level/sspl_ll_d
cdeshmukh: I think it refers to Software as the source or type of IEM being generated

low-level/sspl_ll_d
cdeshmukh: Ok, then its better to have try-exception block in initialize_iem() as @gargpalak048 had done, but do we need 'raise' again at line 727? As we are handling the exception there already and we dont want SSPL to fail out even after catching that exception
@thavanathan

thavanathan: Yeah. we need to remove the `raise` stmt from `initialize_iem` to continue without stopping SSPL. Also there we catched only `EventMessageError`. I believe catching `EventMessageError` is enough to catch initialization failure of IEM framework. Otherwise we need to catch the common `Exception` class instead of `EventMessageError`

cdeshmukh: >  I believe catching EventMessageError is enough to catch initialization failure of IEM framework.

Yes i agree to this

low-level/framework/utils/iem.py
gargpalak048: done

low-level/sspl_ll_d
gargpalak048: > I think it refers to Software as the source or type of IEM being generated

yes

low-level/framework/utils/iem.py
gargpalak048: Ack

low-level/sspl_ll_d
gargpalak048: Agree. Removed the raise statement

unittests/test_iem.py
gargpalak048: done

unittests/test_iem.py
gargpalak048: Removed this piece of code.
No need to remove/install ipmitool.
To mock the ipmitool error, simply call iem_fault() with param = "IPMITOOL_ERROR"
This will generate error & send it to Message Bus

unittests/test_iem.py
gargpalak048: done

unittests/test_iem.py
gargpalak048: $python3 -m pytest
.........................                                                                                                                             [100%]
25 passed in 3.19s

could you try with the latest changes?

gargpalak048:
> No blank lines allowed after function docstring (found 1) (D202)

No blank line present, Invalid codacy comment

jjxsg (approved):

jjxsg (approved):

cdeshmukh:
test this please

unittests/framework/utils/test_iem.py
```diff
@@ -0,0 +1,74 @@
...
+# with this program. If not, see <https://www.gnu.org/licenses/>. For any questions
+# about this software or licensing, please email opensource@seagate.com or
+# cortx-questions@seagate.com.
+
+
+import time
+import unittest
+
+from cortx.utils.iem_framework import EventMessage
+from cortx.utils.process import SimpleProcess
```
thavanathan, line 21: Not used.

unittests/framework/utils/test_iem.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class TestIEM(unittest.TestCase):
+
+    def setUp(self):
+        self.module = None
+        self.event_code = None
+        self.EVENT_CODE = {
+        "IPMITOOL_ERROR" : ["0050010001", "ipmitool"],
+        "IPMITOOL_AVAILABLE" : ["0050010002", "ipmitool"]
```
thavanathan, line 32: nit: right indent.

unittests/framework/utils/test_iem.py
```diff
@@ -0,0 +1,74 @@
...
+        while True:
+            msg = EventMessage.receive()
+            if msg is None:
+                break
+
+    def test_iem_fault_alert_receive(self):
+        """Test iem 'ipmitool' fault alert receive."""
+        print("### TestCase: test_iem_fault_alert_receive\n")
+        self.iem.iem_fault("IPMITOOL_ERROR")
+        time.sleep(10)
```
thavanathan, line 47: Why are we waiting for 10 seconds? Are we using the Kafka service in backend here? If so, we are not supposed to do that. Unittest should not be relying on any external running services. Either you can mock the service or move this test to sanity test suite.

thavanathan (commented):
Looking at the nature of the change, I think we don't need a unittest here. We can have a sanity test to make sure we are able to send and subscribe to IEM alerts.

unittests/framework/utils/test_iem.py
gargpalak048: Addressed @thavanathan.

As discussed, any test having dependancy on external service(for eg:- kafka) should be added in sspl_test framework not unittest.
so, added a test to test iem alerts in alerts.pln

unittests/framework/utils/test_iem.py
gargpalak048: removed the iem unittest

unittests/framework/utils/test_iem.py
gargpalak048: After the latest changes, removed this file

thavanathan (approved):

cdeshmukh:
test this please

low-level/sspl_ll_d
```diff
@@ -713,6 +715,14 @@ def send_thread_controller_request(state):
 #     logger.info(f"Received SIGUSR1 signal to set log level to '{log_level}'")
 #     logger.setLevel(log_level)
 
+def initialize_iem():
+    try:
+        EventMessage.init(component='sspl', source='S')
+        logger.info("IEM framework initiated successfully!!!")
+    except EventMessageError as e:
+        logger.error("Failed to initialize IEM Framework."
```
cdeshmukh, line 723: so if Init fails for some reason, SSPL still later tries to generate IEM irrespective of whether init had failed or not. We need to relate this failure ideally, may be through a marker file with sspl pid in it. and retry init just before sending IEM ..

sspl_test/framework/utils/iem.py
```diff
@@ -0,0 +1,108 @@
...
+        event_type = "fault_resolved"
+        self.create_iem_fields(event, severity, event_type)
+
+    @staticmethod
+    def generate_iem(module, event_code, severity, description):
+        """Generate iem and send it to a MessgaeBroker."""
+        try:
+            logger.info(f"Sending IEM alert for module:{module}"
+                        f" and event_code:{event_code}")
+            EventMessage.send(module=module, event_id=event_code,
```
cdeshmukh, line 104: so we are restricting test till IEM send, which is fine too. Any further issue falls under IEM framework scope, where send, subscription & receive test flow should ideally be there. @sachinpunadikar  hope this unit test is covered in IEM framework tests already

low-level/sspl_ll_d
gargpalak048: Agree @cdeshmukh.
so in case of iem init failure, it will create a file at loc:-

> IEM_INIT_FILE = "/var/%s/sspl/data/iem/iem_init" %(PRODUCT_FAMILY)

and iem before sending the alert, will check for this file, if it is present, it will retry to initialize the IEM framework & then will send the alert and remove this file.

sspl_test/framework/utils/iem.py
gargpalak048: in the iem test, we are sending the mock iem alert and testing alert is received or not in the specific format

low-level/framework/utils/iem.py
```diff
@@ -155,6 +153,20 @@ def check_exsisting_fault_iems(self):
...
+    @staticmethod
+    def generate_iem(module, event_code, severity, description):
+        """Generate iem and send it to a MessgaeBroker."""
+        try:
+            logger.info(f"Sending IEM alert for module:{module}"
+                        f" and event_code:{event_code}")
+            # check if IEM Framework initialized,
+            # if not, retry initializing the IEM Frameowork
+            if os.path.exists(IEM_INIT_FILE):
```
cdeshmukh, line 164: We need to link current sspl pid also in this file. Consider init had failed with some previous sspl service run and IEM were never generated as no situation arises or say this file is not cleaned up by running old sspl instance.
And then new SSPL service instance started, for which IEM init was successful actually, still if this file is present, init would be tried, which is not the intention.

Better you refer and link using SSPL pid, like recently we had done for support bundle case, check #425

low-level/framework/utils/iem.py
```diff
@@ -155,6 +153,20 @@ def check_exsisting_fault_iems(self):
...
+                EventMessage.init(component='sspl', source='S')
```
cdeshmukh, line 165: what if init fails here too ? Then log it and return after checking for init failure marker file itself dont proceed for send()
One problem though, see if we can avoid that - this file existence check will happen before every IEM generation, so its a overhead even if Init had passed at first time itself.

low-level/framework/base/sspl_constants.py
```diff
@@ -37,6 +37,7 @@
 CLI_RESOURCE_PATH = "/opt/seagate/%s/sspl/low-level/tests/manual" % (PRODUCT_FAMILY)
 DATA_PATH = "/var/%s/sspl/data/" % (PRODUCT_FAMILY)
 IEM_DATA_PATH = "/var/%s/sspl/data/iem/sspl_iems"  %(PRODUCT_FAMILY)
+IEM_INIT_FILE = "/var/%s/sspl/data/iem/iem_init" %(PRODUCT_FAMILY)
```
cdeshmukh, line 40: This marker file is know init status, but its created only on failure, so why not name it accordingly, init_failure. Same applies to constant name

low-level/framework/utils/iem.py
gargpalak048: Ack.
done the changes

low-level/framework/utils/iem.py
gargpalak048: this code is already is try except block.
if init fails here as well, it will go to exception block and log an error.
send api will not be called in that case.

gargpalak048: rather than checking init_failed file before every send api, should we move this code of class's __init__ method?
in this way, same obj if sends different iem alerts, only once file exists or not will be checked.
@cdeshmukh should I do this way or any better approach suggestions?

low-level/framework/base/sspl_constants.py
gargpalak048: renamed the file as suggested

low-level/sspl_ll_d
```diff
@@ -713,6 +716,26 @@ def send_thread_controller_request(state):
 #     logger.info(f"Received SIGUSR1 signal to set log level to '{log_level}'")
 #     logger.setLevel(log_level)
 
+def initialize_iem():
+    try:
```
cdeshmukh, line 720: Any pre-existing IEM-INIT FAILED file should be cleaned up here, before trying init at SSPL start

low-level/sspl_ll_d
```diff
@@ -713,6 +716,26 @@ def send_thread_controller_request(state):
...
+        logger.info("IEM framework initiated successfully!!!")
+    except EventMessageError as e:
+        # create IEM_INIT_FAILED file and log current sspl pid.
+        create_file(IEM_INIT_FAILED)
+        with open(IEM_INIT_FAILED, 'w') as f:
+            f.write(str(os.getpid()))
+        logger.critical("Failed to initialize IEM Framework."
+                     f"ERROR:{e}")
+
+def create_file(path):
```
cdeshmukh, line 731: such functions should be collected in common utility class, right now keep it under Utility class only. Ideally they need to be under specific classes, though that classification we will do separately

low-level/sspl_ll_d
gargpalak048: done

low-level/sspl_ll_d
gargpalak048: done

cdeshmukh (approved):

cdeshmukh:
test this please

cdeshmukh:
test this please

sandeepsgt (approved):

jjxsg (approved):

PR #412, opened 2021-06-07
EOS-21075 :  Install coverage pkg with sspl-test rpm

sumedhak27:
## Problem Statement:

Coverage setup setup installs coverage pkg while it's execution if not installed, but this is not the appropriate way.
Coverage version is also "latest", it should be changed to tested version "5.5"

## Solution Design:

Coverage python pkg needs to installed with sspl-test rpm 

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?

### DIFF
--- a/low-level/files/opt/seagate/sspl/setup/sspl_test.py
+++ b/low-level/files/opt/seagate/sspl/setup/sspl_test.py
@@ -57,30 +57,34 @@ class SSPLTestCmd:
             "cortx-sspl-test": None
             }
         # python 3rd party package dependency
-        pip3_3ps_packages_dep = {
+        pip3_packages_dep = {
             "Flask": "1.1.1",
             "coverage": "5.5"
             }
+        if not self.coverage_enabled:
+            pip3_packages_dep.pop("coverage")
         # Collect installed python pkg from dependency list.
         installed_pip3_pkgs = {}
         for pkg in pkg_resources.working_set:
-            if pkg.project_name in pip3_3ps_packages_dep:
+            if pkg.project_name in pip3_packages_dep:
                 installed_pip3_pkgs[pkg.project_name] = pkg.version
         # Validate pip3 python pkg with required version.
-        for pkg, version in pip3_3ps_packages_dep.items():
-            if pkg == "coverage" and not self.coverage_enabled:
-                continue
+        for pkg, version in pip3_packages_dep.items():
             if pkg not in installed_pip3_pkgs or \
                installed_pip3_pkgs[pkg] != version:
                 if pkg in installed_pip3_pkgs:
                     cmd = f'pip3 uninstall -y {pkg}'
                     _, err, ret = SimpleProcess(cmd).run()
                     if ret:
-                        logger.exception(err)
+                        logger.exception(
+                            "Failed to uninstall the pip3 pkg: %s(v%s), "
+                            "due to an Error: %s" % (pkg, version, err))
                 cmd = f'pip3 install {pkg}=={version}'
                 _, err, ret = SimpleProcess(cmd).run()
                 if ret:
-                    logger.exception(err)
+                    logger.exception(
+                        "Failed to install the pip3 pkg: %s(v%s), "
+                        "due to an Error: %s" % (pkg, version, err))
             logger.info(f"Ensured Package Dependency: {pkg}(v{version}).")
         # Validate rpm dependencies
         pkg_validator = PkgV()
--- a/low-level/files/opt/seagate/sspl/setup/sspl_test.py
+++ b/low-level/files/opt/seagate/sspl/setup/sspl_test.py
@@ -40,7 +40,7 @@ class SSPLTestCmd:
     def __init__(self, args: list):
         self.args = args
         self.name = "sspl_test"
-        self.plan = "self"
+        self.plan = "sanity"
         self.coverage_enabled = self.args.coverage
 
         self.dbus_service = DbusServiceHandler()
--- a/low-level/files/opt/seagate/sspl/setup/sspl_test.py
+++ b/low-level/files/opt/seagate/sspl/setup/sspl_test.py
@@ -15,6 +15,7 @@
 import shutil
 import os
 import socket
+import pkg_resources
 
 from cortx.utils.process import SimpleProcess
 from cortx.utils.conf_store import Conf
@@ -57,14 +58,24 @@ class SSPLTestCmd:
             }
         # python 3rd party package dependency
         pip3_3ps_packages_test = {
-            "Flask": "1.1.1",
-            "coverage": "5.5"
+            "Flask": "1.1.1"
             }
         pkg_validator = PkgV()
         pkg_validator.validate_pip3_pkgs(host=socket.getfqdn(),
             pkgs=pip3_3ps_packages_test, skip_version_check=False)
         pkg_validator.validate_rpm_pkgs(host=socket.getfqdn(),
             pkgs=rpm_deps, skip_version_check=True)
+        # Validae the coverage==5.5 python pkg installation
+        coverage_pkg = [p for p in pkg_resources.working_set
+                        if 'coverage' == p.project_name]
+        if not coverage_pkg or coverage_pkg[0].version != '5.5':
+            if coverage_pkg:
+                _, err, ret = SimpleProcess('pip3 uninstall -y coverage').run()
+                if ret:
+                    logger.exception(err)
+            _, err, ret = SimpleProcess('pip3 install coverage==5.5').run()
+            if ret:
+                logger.exception(err)
         # Load global, sspl and test configs
         Conf.load(SSPL_CONFIG_INDEX, sspl_config_path)
         Conf.load(SSPL_TEST_CONFIG_INDEX, sspl_test_config_path)
--- a/low-level/files/opt/seagate/sspl/setup/sspl_test.py
+++ b/low-level/files/opt/seagate/sspl/setup/sspl_test.py
@@ -15,7 +15,7 @@
 import shutil
 import os
 import socket
-import pkg_resources
+from pkg_resources import Requirement, working_set, VersionConflict
 
 from cortx.utils.process import SimpleProcess
 from cortx.utils.conf_store import Conf
@@ -63,29 +63,36 @@ class SSPLTestCmd:
             }
         if not self.coverage_enabled:
             pip3_packages_dep.pop("coverage")
-        # Collect installed python pkg from dependency list.
-        installed_pip3_pkgs = {}
-        for pkg in pkg_resources.working_set:
-            if pkg.project_name in pip3_packages_dep:
-                installed_pip3_pkgs[pkg.project_name] = pkg.version
+
         # Validate pip3 python pkg with required version.
         for pkg, version in pip3_packages_dep.items():
-            if pkg not in installed_pip3_pkgs or \
-               installed_pip3_pkgs[pkg] != version:
-                if pkg in installed_pip3_pkgs:
-                    cmd = f'pip3 uninstall -y {pkg}'
-                    _, err, ret = SimpleProcess(cmd).run()
-                    if ret:
-                        logger.exception(
-                            "Failed to uninstall the pip3 pkg: %s(v%s), "
-                            "due to an Error: %s" % (pkg, version, err))
+            installed_pkg = None
+            uninstalled_pkg = False
+            try:
+                pkg_req = Requirement.parse(f"{pkg}=={version}")
+                installed_pkg = working_set.find(pkg_req)
+            except VersionConflict:
+                cmd = f'pip3 uninstall -y {pkg}'
+                _, err, ret = SimpleProcess(cmd).run()
+                if ret:
+                    raise TestException(
+                        "Failed to uninstall the pip3 pkg: %s(v%s), "
+                        "due to an Error: %s" % (pkg, version, err))
+                uninstalled_pkg = True
+            except Exception as err:
+                raise TestException(
+                    "Failed at verification of pip3 pkg: %s, "
+                    "due to an Error: %s" % (pkg, err))
+
+            if not installed_pkg or uninstalled_pkg:
                 cmd = f'pip3 install {pkg}=={version}'
                 _, err, ret = SimpleProcess(cmd).run()
                 if ret:
-                    logger.exception(
+                    raise TestException(
                         "Failed to install the pip3 pkg: %s(v%s), "
                         "due to an Error: %s" % (pkg, version, err))
             logger.info(f"Ensured Package Dependency: {pkg}(v{version}).")
+
         # Validate rpm dependencies
         pkg_validator = PkgV()
         pkg_validator.validate_rpm_pkgs(
--- a/low-level/files/opt/seagate/sspl/setup/sspl_test.py
+++ b/low-level/files/opt/seagate/sspl/setup/sspl_test.py
@@ -41,7 +41,7 @@ class SSPLTestCmd:
         self.args = args
         self.name = "sspl_test"
         self.plan = "self"
-        self.coverage_enabled = False
+        self.coverage_enabled = self.args.coverage
 
         self.dbus_service = DbusServiceHandler()
         if args.config and args.config[0]:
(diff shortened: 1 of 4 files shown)


sspl_test/sspl-test.spec
```diff
@@ -78,6 +78,18 @@ else
     #touch ${SSPL_DIR}/sspl_test/keep_flask
 fi
 
+# Check and install required coverage version
+cov=`pip3.6 freeze | grep coverage`
+if [[ -n $cov ]]; then
+    ver=${cov##*=}
+    if [[ "$ver" != "5.5" ]]; then
+        pip3 uninstall -y coverage
+        pip3 install coverage==5.5
```
cdeshmukh, line 87: coverage pkg needs to be part of cortx-prerequisites pkg maintained by RE. 

@ujjwalpl & @shailesh-vaidya cortx-sspl-test needs this 'coverage' pkg in dev/test env where code coverage needs to generated & collected, is it ok to keep it cortx-prerequisites, even if it is needed in dev/qa env only ?

cdeshmukh: @sumedhak27 follow up with Shailesh tomorrow

sumedhak27: okay sure..

shailesh-vaidya: @cdeshmukh @sumedhak27  I guess coverage package is build time dependency and not run time dependency (Not required for deployment) . We should only add run time dependencies in `cortx-prereq` packages to keep third party packages less. 
You can use `coverage` package during build time without any issues. Let me know if you face any problem there.

cdeshmukh: @shailesh-vaidya Its cortx-sspl-test rpm post install requirement, its not bundled in rpm during build. 
Now cortx-sspl-test itself is optional, so it doesnt affect deployment. Though for running sspl self-tests after deployment this cortx-sspl-test package is needed. 
At run time coverage module searched for, if tests started with argument to generate code coverage.

So i still see it as run time dependency, unless am thinking something wrong, please suggest.

cdeshmukh: @sumedhak27 Spoke to @ujjwalpl about this. First it doesnt need to be part of cortx-prerequisites given its not needed by a must cortx pkg. Since needed by optional sspl-test package, its deployment should relate to that.

Secondly, any dependent package installation should happen through 'Requires' attribute if dependent pkg is an rpm. Else such installation should not happen through rpm post-install stage, but through mini provisioner interface. So this coverage & even flask package installation should happen during mini-provisioner 'test' invocation, basically check for pkg & install it. This would also mean, any test execution should happen through mini provisioner interface only going forward & not through direct sspl test scripts like its done today.

cdeshmukh: THis was discussed again. 
1. (preferred) If not possible to pull dependent pkg, through rpm 'Requires' attribute, manually install dependent packages before sspl-test rpm installation & update the instructions in wiki with these steps.
Mini provisioner test interface should check for these dependent pkg installation in validate() and fail listing these pkgs.

2. Check & Install dependent packages in mini provisioner interface 'test' before tests are performed even
@sumedhak27

sumedhak27: so If validator test fails in 'sspl_setup test' for coverage, then should we install the pkg?

sumedhak27: added Validation and installation of coverage.py in sspl_test mini provisioner's validate phase.

sspl_test/coverage/coverage_setup.py
```diff
@@ -170,10 +165,11 @@ def coverage_reset():
 
 def print_help():
     print('Error: Incorrect arguments to coverage_setup file.\n'
-        'cmd : python3 coverage_setup.py [start/stop]\n'
-        'Args => \n'
-        '    start : Set-up the environment for code coverage.\n'
-        '    stop : Reset the sspl environmet to normal.')
+          'cmd : python3 coverage_setup.py [start/stop]\n'
```
sumedhak27, line 168: flake8 relate change.

sspl_test/coverage/coverage_setup.py
```diff
@@ -186,4 +182,4 @@ def print_help():
         coverage_reset()
     else:
         print_help()
-        sys.exit(100)
\ No newline at end of file
+        sys.exit(100)
```
sumedhak27, line 185: flake8 related change.

sspl_test/coverage/coverage_setup.py
```diff
@@ -147,8 +141,9 @@ def coverage_reset():
         modification_time = os.path.getmtime(REPORT_PATH)
 
         if (time.time() - modification_time) < 100:
-            modification_time = time.strftime('%Y-%m-%d %H:%M:%S',
-                                        time.localtime(modification_time))
+            modification_time = \
+                time.strftime('%Y-%m-%d %H:%M:%S',
+                              time.localtime(modification_time))
```
sumedhak27, line 146: flake8 related change.

sspl_test/coverage/coverage_setup.py
```diff
@@ -170,10 +165,11 @@ def coverage_reset():
...
     print('Error: Incorrect arguments to coverage_setup file.\n'
-        'cmd : python3 coverage_setup.py [start/stop]\n'
-        'Args => \n'
-        '    start : Set-up the environment for code coverage.\n'
-        '    stop : Reset the sspl environmet to normal.')
+          'cmd : python3 coverage_setup.py [start/stop]\n'
+          'Args => \n'
+          '    start : Set-up the environment for code coverage.\n'
+          '    stop : Reset the sspl environmet to normal.')
+
```
sumedhak27, line 172: Flake8 related change.

sspl_test/coverage/coverage_setup.py
```diff
@@ -57,31 +57,25 @@ def generate_cov_report(signal_number, frame):
         outfile='%scoverage/sspl_xml_coverage_report.xml')
     logger.info(f'XML coverage report generated with coverage of {cov_per} percentage.')
     ## Enable below code to inable HTML report generation
- 	# html_cov_per = co.html_report(
- 	#                     directory='%scoverage/sspl_html_coverage',
- 	#                     ignore_errors=True,
- 	#                 )
- 	# logger.info(f'HTML coverage report geverated with coverage of {html_cov_per} percentage.')\
-"""%(DATA_PATH, DATA_PATH)
+    # html_cov_per = co.html_report(
```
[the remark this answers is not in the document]
sumedhak27: There were few flake8 issues in this file, so I thought of correcting them.
But sure will mark it in this PR so that you won't waste time in understanding the change.
Can remove them as well if we really don't want them in this PR.

low-level/files/opt/seagate/sspl/setup/sspl_test.py
```diff
@@ -64,6 +65,17 @@ def validate(self):
             pkgs=pip3_3ps_packages_test, skip_version_check=False)
         pkg_validator.validate_rpm_pkgs(host=socket.getfqdn(),
             pkgs=rpm_deps, skip_version_check=True)
+        # Validae the coverage==5.5 python pkg installation
```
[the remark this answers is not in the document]
sumedhak27: Addressed. 👍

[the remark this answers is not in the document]
sumedhak27: * Handled for flask (made it generic). 
* If pkg is not installed or the required version is not installed, we are installing the exact version with this change. please review.
* sspl test interface will log an error/exception if any during installation and exit out.

sumedhak27:
commenting to prevent auto close.

low-level/files/opt/seagate/sspl/setup/sspl_test.py
```diff
@@ -56,14 +57,35 @@ def validate(self):
             "cortx-sspl-test": None
             }
         # python 3rd party package dependency
-        pip3_3ps_packages_test = {
-            "Flask": "1.1.1"
+        pip3_3ps_packages_dep = {
```
cdeshmukh, line 60: Just keep it as pip3_packages_dep

sspl_test/coverage/coverage_setup.py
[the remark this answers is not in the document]
cdeshmukh: in such cases, mention about flake8 upfront as review comment urself, so we can relate faster

low-level/files/opt/seagate/sspl/setup/sspl_test.py
```diff
@@ -56,14 +57,35 @@ def validate(self):
...
+            "Flask": "1.1.1",
+            "coverage": "5.5"
             }
+        # Collect installed python pkg from dependency list.
+        installed_pip3_pkgs = {}
+        for pkg in pkg_resources.working_set:
+            if pkg.project_name in pip3_3ps_packages_dep:
+                installed_pip3_pkgs[pkg.project_name] = pkg.version
+        # Validate pip3 python pkg with required version.
+        for pkg, version in pip3_3ps_packages_dep.items():
```
cdeshmukh, line 70: why two loops, can be done in one loop

low-level/files/opt/seagate/sspl/setup/sspl_test.py
```diff
@@ -56,14 +57,35 @@ def validate(self):
...
+            if pkg == "coverage" and not self.coverage_enabled:
```
cdeshmukh, line 71: use this condition to above before line 64 to remove coverage from pip3_3ps_packages_dep  dict itself, so in for loop we dont deal with any specifics

low-level/files/opt/seagate/sspl/setup/sspl_test.py
```diff
@@ -56,14 +57,35 @@ def validate(self):
...
+                continue
+            if pkg not in installed_pip3_pkgs or \
+               installed_pip3_pkgs[pkg] != version:
+                if pkg in installed_pip3_pkgs:
+                    cmd = f'pip3 uninstall -y {pkg}'
```
cdeshmukh, line 76: Wont this need permission would be needed to install pkgs? Confirm with provisioner that they would invoke this with user with enough permissions. Also if this installation fails, gracefully exit from sspl_test itself, wth clear message, just following wont be enough
> logger.exception(err)

cdeshmukh (changes requested):

low-level/files/opt/seagate/sspl/setup/sspl_test.py
sumedhak27: In the first loop, I am getting all the pip pkgs installed on the node, and involving it in the list only if we are depending on it.  so we get installed_dep
In the second loop, I am checking if we have (dep - installed_dep) as non-zero then install those pkgs, also if the version number of installed_dep does not match then install the correct version.

How can we do it in a single loop?? I have no clue.
Another approach is still taking two loops.
In the first loop while retrieving the installed pkg we can check if that pkg is there in dep if it matches the version we can remove it from the dep list,  even then, we would need a second loop to go over pkg that is remaining in dep list.

sspl_test/coverage/coverage_setup.py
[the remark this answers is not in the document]
sumedhak27: sure 👍

low-level/files/opt/seagate/sspl/setup/sspl_test.py
sumedhak27: addressed. 👍

low-level/files/opt/seagate/sspl/setup/sspl_test.py
sumedhak27: addressed 👍

low-level/files/opt/seagate/sspl/setup/sspl_test.py
jjxsg: You are traversing through every installed package and then comparing if any of them matches with the packages you have in dependency list. Ideally it should be other way around.

low-level/files/opt/seagate/sspl/setup/sspl_test.py
```diff
@@ -56,14 +57,39 @@ def validate(self):
...
+        installed_pip3_pkgs = {}
+        for pkg in pkg_resources.working_set:
+            if pkg.project_name in pip3_packages_dep:
+                installed_pip3_pkgs[pkg.project_name] = pkg.version
+        # Validate pip3 python pkg with required version.
+        for pkg, version in pip3_packages_dep.items():
+            if pkg not in installed_pip3_pkgs or \
+               installed_pip3_pkgs[pkg] != version:
+                if pkg in installed_pip3_pkgs:
+                    cmd = f'pip3 uninstall -y {pkg}'
```
jjxsg, line 76: You can use [Requirement object](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#requirement-objects) from `pkg_resources`. Build a requirement string with name and version of a package and then you can use a function like this:
```
def check_package_exists(package):
    req = pkg_resources.Requirement.parse(package)
    return any(dist in req for dist in pkg_resources.working_set)
```
```
$ pip freeze | grep pytest
pytest==6.2.2

>>> check_package_exists('pytest==6.2.3')
False

>>> check_package_exists('pytest==6.2.2')
True
```

sumedhak27: Thats great @jjxsg , I was also checking how to use Requirement obj but failed to find any example.
This is an improvement but still we can looping through the working_set, There is one method that I see that can prevent this looping, but I failed to find how to use it can you help?
```
find(req)
Find a distribution matching req (a Requirement instance). 
If there is an active distribution for the requested project,
this returns it, as long as it meets the version requirement specified by req. 
But, if there is an active distribution for the project and it does not 
meet the req requirement, VersionConflict is raised. If there is no active 
distribution for the requested project, None is returned.
```

jjxsg: That was my initial implementation, but it will require handling the exception and probably `find function` will be using similar loop behind the scenes. You can explore more, here is basic implementation:
```
def check_package_exists(package):
    result = None
    req = pkg_resources.Requirement.parse(package)
    try:
        result = pkg_resources.working_set.find(req)
    except VersionConflict:
        print("Version Conflict")
    return (result is not None)
```
```
>>> check_package_exists('pytest==6.2.2')
True
>>> check_package_exists('pytest==6.2.3')
Version Conflict
False
>>> check_package_exists('abcd==6.2.3')
False

File: low-level/files/opt/seagate/sspl/setup/sspl_test.py (hunk 9 above)
Comment (sumedhak27) [reply low-level/files/opt/seagate/sspl/setup/sspl_test.py]: @jjxg, the working set is bigger so traversing through the installed pkg and then for each installed pkg checking if it is there in small set of dependency is better as compared to for each dependency pkg traverse big installed pkg list.

File: low-level/files/opt/seagate/sspl/setup/sspl_test.py (hunk 12 above)
Comment (sumedhak27) [reply low-level/files/opt/seagate/sspl/setup/sspl_test.py:76]: this looks great, Thanks 👍

Comment (sumedhak27) [reply low-level/files/opt/seagate/sspl/setup/sspl_test.py]: addressed 👍

File: low-level/files/opt/seagate/sspl/setup/sspl_test.py (hunk 9 above)
Comment (sumedhak27) [reply low-level/files/opt/seagate/sspl/setup/sspl_test.py]: addressed 👍

File: low-level/files/opt/seagate/sspl/setup/sspl_test.py [hunk 13]
```diff
@@ -56,14 +57,46 @@ def validate(self):
             "cortx-sspl-test": None
             }
         # python 3rd party package dependency
-        pip3_3ps_packages_test = {
-            "Flask": "1.1.1"
+        pip3_packages_dep = {
```
jjxsg, line 60: Rename `pip3_packages_dep` to `python_packages_dep` or `python3_packages_dep`

low-level/files/opt/seagate/sspl/setup/sspl_test.py
jjxsg: Looks fine.

sspl_test/coverage/coverage_setup.py
```diff
@@ -57,31 +57,25 @@ def generate_cov_report(signal_number, frame):
...
+""" % (DATA_PATH, DATA_PATH)
 
 PATCH_4 = """\
     signal.signal(signal.SIGUSR1, generate_cov_report)\
 """
 
+
 def coverage_setup():
-    """Installs pip3 coverage pkg and creates required files for code coverage.
+    """Creates required files for code coverage.
```
jjxsg, line 73:
```suggestion
    """
    Creates required files for code coverage.
```

low-level/files/opt/seagate/sspl/setup/sspl_test.py
sumedhak27: okay @jjxsg 👍 , @cdeshmukh agree with `python3_package_dep`?

sspl_test/coverage/coverage_setup.py
sumedhak27: addressed, Now throws codacy issue to start at first line 🤣

jjxsg (approved):

sspl_test/coverage/coverage_setup.py
jjxsg: Yes, we need to get one of them disabled.

low-level/files/opt/seagate/sspl/setup/sspl_test.py
```diff
@@ -40,7 +41,7 @@ def __init__(self, args: list):
         self.args = args
         self.name = "sspl_test"
         self.plan = "self"
```
shriya-deshmukh, line 43: Change self.plan = "sanity" as we have renamed "self" plan to "sanity".

sumedhak27: addressed 👍

shriya-deshmukh (approved):

low-level/files/opt/seagate/sspl/setup/sspl_test.py
cdeshmukh: yes, makes sense

cdeshmukh (approved):

sspl_test/coverage/coverage_setup.py
cdeshmukh: Some codacy issues keep conflicting

PR #441, opened 2021-06-29
EOS-20921: Fix for Bug- Setup log is not created and error info lost …

gargpalak048:
…during post_install failure

Setup log is not created and error info lost during post_install failure

sspl log dir used to get configured in post_install process stage, so any error in sspl-setup validation stage was not getting logged. Moved the sspl-setup log dir configuration to the post_install Validation stage.

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?


cortx-admin:

low-level/files/opt/seagate/sspl/setup/sspl_post_install.py
```diff
@@ -60,6 +60,9 @@ def validate(self):
         """
         machine_id = Utility.get_machine_id()
 
+        # configure sspl-setup log dir before validation stage.
```
cdeshmukh, line 63: well why not in init itself.
validate() should have validations alone and not any pre-requisite prep

cdeshmukh (commented):
And be extra cautious while dealing with these mini prov interfaces, as it affects deployment. make sure you have tested this enough

low-level/files/opt/seagate/sspl/setup/sspl_post_install.py
gargpalak048: Everytime PostInstall  class object will be created, sspl-setup dir will be configured if put in init.

gargpalak048:
> And be extra cautious while dealing with these mini prov interfaces, as it affects deployment. make sure you have tested this enough

we have just moved the sspl-setup log configuration from process stage to validation stage, I don't think it will have any impact on deployment.
I have tested thoroughly on dev VM
Need HW testing as well?

low-level/files/opt/seagate/sspl/setup/sspl_post_install.py
cdeshmukh: check for dir existence and if missing, then only create. validate is not right place for such actions

gargpalak048: Addressed

gargpalak048: @cdeshmukh we need to configure sspl-setup log dir in validate stage only.
because while configuring it requires values in product name etc which are getting fetched in validation stage only.
And these values cannot be fetched in init stage because in sspl-setup, Conf index gets loaded in validation stage only
so after the validation of product name & setup then we can configure the setup-log dir

sandeepsgt:
Instead all of this , you can just use RotatingFileHandler in init_logging function 
Check this [line](https://github.com/Seagate/cortx-monitor/blob/baabce3155d8af3f361603c9c637af9e772cfa36/low-level/files/opt/seagate/sspl/setup/setup_logger.py#L30), 
`    handler = logging.handlers.RotatingFileHandler(<logfile_location>, mode='a', maxBytes=2000000, backupCount=5)`

cdeshmukh:
test this please

gargpalak048:
> Instead all of this , you can just use RotatingFileHandler in init_logging function
> Check this [line](https://github.com/Seagate/cortx-monitor/blob/baabce3155d8af3f361603c9c637af9e772cfa36/low-level/files/opt/seagate/sspl/setup/setup_logger.py#L30),
> ` handler = logging.handlers.RotatingFileHandler(<logfile_location>, mode='a', maxBytes=2000000, backupCount=5)`

Ack @sandeepsgt
Earlier sysloghandler was implemented to keep the aligned with other logging modules.
But as sspl-setup scripts will always run before the service starts, there is no such requirement to write logs to syslog first and can directly use the Filehandler to dump logs to file path.
It will decrease the complexity of code as well.
I have incorporated the changes in the latest commit.
@cdeshmukh @sandeepsgt could you review the latest changes

low-level/files/opt/seagate/sspl/setup/setup_logger.py
```diff
@@ -27,11 +29,17 @@ def init_logging(syslog_host="localhost", syslog_port=514, console_output=False)
     """Initialize logging for sspl-setup."""
     _logger.setLevel(logging.INFO)
     # set Logging Handlers
-    handler = logging.handlers.SysLogHandler(
-                address=(syslog_host, syslog_port))
-    syslog_format = "%(name)s[%(process)d]: " \
-                    "%(levelname)s %(message)s (%(filename)s:%(lineno)d)"
-    formatter = logging.Formatter(syslog_format)
+    Utility().create_file(SETUP_LOG_PATH)
```
sandeepsgt, line 32: Is file creation is really required ? will file not get created when you write log?

gargpalak048: @sandeepsgt  on the fresh VM, /var/log/cortx  dir won't exist. So need to configure the dir for the first time.
create_file() will check for complete path and create file if not exist.

satish-seagate: @gargpalak048 Can you please confirm that created directory/file have sspl-ll user permission?

gargpalak048: #ls -lrt  /var/log/cortx/sspl/sspl-setup.log
**-rw-r--r--** 1 root root 1031 Jul 12 02:16 /var/log/cortx/sspl/sspl-setup.log

file has 644 permissions @satish-seagate

sandeepsgt: Whichever user runs miniprovisioner interfaces, file will be created with that user's permission.
Correct me if I am wrong, but I don't think this file need to have sspl-ll:sspl-ll user permission.

satish-seagate: @sandeepsgt may I know why permission is not required?

sandeepsgt: `Whichever user runs miniprovisioner interfaces, file will be created with that user's permission.`
SSPL service has nothing to do with this file. So. I think permission is not required.

satish-seagate: Okay. @gargpalak048 @sandeepsgt I am okay with the change then.

sandeepsgt (approved):

satish-seagate (approved):

jjxsg (approved):

low-level/files/opt/seagate/sspl/setup/setup_logger.py
cdeshmukh: Better check if logs seen created properly in system where full Cortx Stck is deployed.

First dir /var/log/cortx creation should/would be done by Provisioner during deployment itself, though we should call that out as pre-requisite to provisioner if they are not doing it before component setup scripts are invoked.

Secondly dir needs to have write permission for other users too, beyond user creating that dir. Otherwise how sspl-ll user would be able to create sspl-setup log file then, check dir permission too here.

low-level/files/opt/seagate/sspl/setup/setup_logger.py
```diff
@@ -27,11 +29,17 @@ def init_logging(syslog_host="localhost", syslog_port=514, console_output=False)
...
+
+    handler = logging.handlers.RotatingFileHandler(
```
cdeshmukh, line 34: Ok this is enabled for sspl-setup logs only, that way we are doing away with additional config file also. But we will loose flexibility to have config params for this, like maxBytes & backup count here. 

Can you put these configs in sspl.conf so we can use from there.

If this is seen stable, same way we can adopt for sspl & other logs also

gargpalak048:
> First dir /var/log/cortx creation should/would be done by Provisioner during deployment itself, though we should call that out as pre-requisite to provisioner if they are not doing it before component setup scripts are invoked.

yes Chetan.  /var/log/cortx is created by provisioner  during deployment.
and under this dir, sspl dir is created by monitor.

gargpalak048:
> Secondly dir needs to have write permission for other users too, beyond user creating that dir. Otherwise how sspl-ll user would be able to create sspl-setup log file then, check dir permission too here.

 ls -lrt /var/log/cortx/
total 8
drwxr-xr-x 3 root root 4096 Jul 12 23:17 utils
drwxr-xr-x 2 root root 4096 Jul 12 23:17 sspl

sspl dir have correct permissions

low-level/files/opt/seagate/sspl/setup/setup_logger.py
gargpalak048: > Can you put these configs in sspl.conf so we can use from there.

 @sandeepsgt @cdeshmukh
Then we will land up in the same problem as earlier,
 as sspl.conf is being created & loaded by post-install process stage.
 and here init_logging function is being called by sspl-setup script, before post-install interface.

gargpalak048: can we keep these conf values in sspl_constant.py instead?

sandeepsgt: I am fine if values are not configurable for this. Because I feel there we don't need to change this once code is deployed. Also, SSPL setup logs will be very small in amount. So, current values should be fine
Just FYI - Earlier also it was [not configurable](https://github.com/Seagate/cortx-monitor/blob/main/low-level/files/etc/logrotate.d/sspl_setup_logs)

low-level/files/opt/seagate/sspl/setup/setup_logger.py
sandeepsgt: If /var/log/cortx/sspl/ directory is not present, then it will be created with `drwxr-xr-x 2 root root` permission.
sspl_log file may not have permission issue with this, because sspl_log is created by rsyslog, not SSPL.

sandeepsgt (commented):
Rebase needed

sandeepsgt (approved):

low-level/files/opt/seagate/sspl/setup/setup_logger.py
cdeshmukh: I believe write permission for dir & file would already exists, including for sspl-ll user, otherwise it wouldnt be working today also. Though test it & confirm for sure

low-level/files/opt/seagate/sspl/setup/setup_logger.py
cdeshmukh: Ok, i think let this be, given these are setup logs only. how often setup would be executed on a deployed system anyways. Current number suffice, including rotation

cdeshmukh (approved):

cdeshmukh:
test this please

cdeshmukh:
test this please

PR #251, opened 2020-11-26
EOS-10911 Move all configs except common configs out of provisioner sspl.sls

shubhamseagate:
…o /etc/sspl.conf

## Problem Statement
<pre>
  <code>
    Story Ref (if any):
    Please add Problem statement here...
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Please add Problem decription here...
  </code>
</pre>
## Solution
<pre>
  <code>
    Please add summary of the change,List any dependencies that are required for this change.
  </code>
</pre>
## Sanity testing on RPM done
<pre>
  <code>
    - [ ] Yes
    - [ ] No
  </code>
</pre>
## Unit/Manual Testing Description
<pre>
  <code>
    Please describe the tests that you ran to verify your changes.
  </code>
</pre>

PR #444, opened 2021-07-02
EOS-22059: Fault description for activating, deactivating alerts does…

shriya-deshmukh:
…n't include the details related to time.

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?

### DIFF
--- a/low-level/sensors/impl/centos_7/service_monitor.py
+++ b/low-level/sensors/impl/centos_7/service_monitor.py
@@ -187,20 +187,10 @@ class Service:
 
     @property
     def is_enabled(self):
-        path_array = self.properties_iface.Get(SERVICE_IFACE, 'ExecStart')
-        # check if service is configured on system,
-        # if yes then check if it is in enabled state.
-        try:
-            command_line_path = str(path_array[0][0])
-        except IndexError as err:
-            logger.error(
-                "Unable to find %s path due to %s" % (self.name, err))
-            command_line_path = "NA"
-        if command_line_path != "NA":
+        if self.properties_iface.Get(UNIT_IFACE, 'UnitFileState'):
             return 'disabled' not in self.properties_iface.Get(
                 UNIT_IFACE, 'UnitFileState')
-        else:
-            return False
+        return False
 
     def is_nonactive_for_threshold_time(self):
         return time.time() - self.nonactive_enter_timestamp > self.nonactive_threshold
--- a/low-level/sensors/impl/centos_7/service_monitor.py
+++ b/low-level/sensors/impl/centos_7/service_monitor.py
@@ -187,8 +187,20 @@ class Service:
 
     @property
     def is_enabled(self):
-        return 'enabled' in self.properties_iface.Get(
-            UNIT_IFACE, 'UnitFileState')
+        path_array = self.properties_iface.Get(SERVICE_IFACE, 'ExecStart')
+        # check if service is configured on system,
+        # if yes then check if it is in enabled state.
+        try:
+            command_line_path = str(path_array[0][0])
+        except IndexError as err:
+            logger.error(
+                "Unable to find %s path due to %s" % (self.name, err))
+            command_line_path = "NA"
+        if command_line_path != "NA":
+            return 'disabled' not in self.properties_iface.Get(
+                UNIT_IFACE, 'UnitFileState')
+        else:
+            return False
 
     def is_nonactive_for_threshold_time(self):
         return time.time() - self.nonactive_enter_timestamp > self.nonactive_threshold
--- a/low-level/sensors/impl/centos_7/service_monitor.py
+++ b/low-level/sensors/impl/centos_7/service_monitor.py
@@ -446,7 +446,7 @@ class ServiceMonitor(SensorThread, InternalMsgQ):
             if self.services[service].is_nonactive_for_threshold_time():
                 self.services[service].new_service_state(FailedState)
                 self.raise_alert(self.get_alert(self.services[service],
-                                                FailedAlert))
+                                                InactiveAlert))
 
     def raise_iem(self, service, alert_type):
         """Raise iem alert for kafka service."""
--- a/low-level/sensors/impl/centos_7/service_monitor.py
+++ b/low-level/sensors/impl/centos_7/service_monitor.py
@@ -54,7 +54,7 @@ FailedAlert = Alert("fault", "{} in {} state.", "{} service is unavailable.",
 InactiveAlert = Alert("fault", "{} in {} state for more than {} seconds.",
                       "{} service is unavailable.",
                       "Try to restart the service")
-ResolvedAlert = Alert("fault_resolved", "{} in {} state.",
+ResolvedAlert = Alert("fault_resolved", "{} in {} state from last {} seconds.",
                       "{} service is available now.",
                       "No action required")
 
@@ -67,6 +67,7 @@ class ActiveState:
 
     @staticmethod
     def enter(service):
+        service.active_enter_timestamp = time.time()
         Service.non_active.discard(service.name)
 
 
@@ -152,6 +153,9 @@ class Service:
 
     alerts = Queue()
     non_active = set()
+    # once fault resolved detected for service, add service name
+    # into the active_services list.
+    active_services = set()
     monitoring_disabled = set()
 
     def __init__(self, unit):
@@ -169,21 +173,28 @@ class Service:
         self.previous_substate = "N/A"
         self.previous_pid = "N/A"
         self.nonactive_enter_timestamp = time.time()
+        self.active_enter_timestamp = time.time()
         self.nonactive_threshold = int(
             Conf.get(SSPL_CONF,
                      f"{ServiceMonitor.name().upper()}>threshold_inactive_time",
                      '60'))
+        self.active_threshold = int(Conf.get(SSPL_CONF,
+                     f"{ServiceMonitor.name().upper()}>threshold_active_time",
+                     '30'))
         self.properties_changed_signal = None
         self._service_state = ActiveState
         self._unit_state = None
 
     @property
     def is_enabled(self):
-        return 'disabled' not in self.properties_iface.Get(
+        return 'enabled' in self.properties_iface.Get(
             UNIT_IFACE, 'UnitFileState')
 
     def is_nonactive_for_threshold_time(self):
         return time.time() - self.nonactive_enter_timestamp > self.nonactive_threshold
+
+    def is_active_for_threshold_time(self):
+        return time.time() - self.active_enter_timestamp > self.active_threshold
 
     def new_service_state(self, new_state):
         if self._service_state != new_state:
@@ -216,9 +227,8 @@ class Service:
 
         if self.state != self.previous_state:
             if self.state == "active":
-                if self._service_state is FailedState:
-                    Service.alerts.put(
-                        ServiceMonitor.get_alert(self, ResolvedAlert))
+                if self._service_state is FailedState and self.name not in Service.active_services:
+                    Service.active_services.add(self.name)
                 self.new_service_state(ActiveState)
             elif self.state == "failed":
                 if self._service_state is not FailedState:
@@ -248,6 +258,7 @@ class Service:
         service.new_service_state(data["service_monitor_state"])
         service.state = data["service_state"]
         service.nonactive_enter_timestamp = data["nonactive_enter_timestamp"]
+        service.active_enter_timestamp = data["active_enter_timestamp"]
         return service
 
     def dump_to_cache(self):
@@ -257,7 +268,8 @@ class Service:
         data = {
             "service_state": self.state,
             "service_monitor_state": self._service_state,
-            "nonactive_enter_timestamp": self.nonactive_enter_timestamp
+            "nonactive_enter_timestamp": self.nonactive_enter_timestamp,
+            "active_enter_timestamp": self.active_enter_timestamp
         }
         store.put(data, f"{CACHE_PATH}/{self.name}")
 
@@ -362,6 +374,7 @@ class ServiceMonitor(SensorThread, InternalMsgQ):
                         self.services[service].new_unit_state(EnabledState)
                     # Check for services in intermediate state(not active)
                     self.check_nonactive_services()
+                    self.check_active_services()
                 time.sleep(self.thread_sleep)
                 iterations += 1
             logger.info("ServiceMonitor gracefully breaking out " +
@@ -448,6 +461,19 @@ class ServiceMonitor(SensorThread, InternalMsgQ):
                 self.raise_alert(self.get_alert(self.services[service],
                                                 InactiveAlert))
 
+    def check_active_services(self):
+        """
+            Monitor active services.
+
+            Raise fault resolved alert for active services
+            if service stays in active state for threshhold time.
+        """
+        for service in Service.active_services.copy():
+            if self.services[service].is_active_for_threshold_time():
+                self.raise_alert(
+                    self.get_alert(self.services[service], ResolvedAlert))
+                Service.active_services.discard(service)
+
     def raise_iem(self, service, alert_type):
         """Raise iem alert for kafka service."""
         if service == "kafka.service" and alert_type == "fault":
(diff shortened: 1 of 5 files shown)


sumedhak27 (approved):

sandeepsgt (approved):

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -446,7 +446,7 @@ def check_nonactive_services(self):
             if self.services[service].is_nonactive_for_threshold_time():
                 self.services[service].new_service_state(FailedState)
                 self.raise_alert(self.get_alert(self.services[service],
-                                                FailedAlert))
+                                                InactiveAlert))
```
cdeshmukh, line 449: I believe the bug is about updating description, to reflect that SSPL waited for configured time (180 seconds) before raising the alerts !? Why change in service alert type here

shriya-deshmukh: FailedAlert, InactiveAlert its an namedtuple, 
Here we have to pass InactiveALert in this tuple we have defined the description as - eg. "statsd.service in inactive state for more than 180 seconds."

FailedAlert = Alert("fault", "{} in {} state.", "{} service is unavailable.",
                    "Try to restart the service")
InactiveAlert = Alert("fault", "{} in {} state for more than {} seconds.",
                      "{} service is unavailable.",
                      "Try to restart the service")
ResolvedAlert = Alert("fault_resolved", "{} in {} state.",
                      "{} service is available now.",
                      "No action required")

thavanathan (approved):

low-level/sensors/impl/centos_7/service_monitor.py
cdeshmukh: >Stable R2: #251: Fault description for activating, deactivating alerts doesn't include the details related to time for which it has waited before reporting the alert.

Before raising fault-resolved alert, logic should wait n see if service stays recovered for sometime and add in description about this time wait too. 

>ResolvedAlert = Alert("fault_resolved", "{} in {} state **from last {} seconds.**",
"{} service is available now.",
"No action required")


Am i missing something ?

sumedhak27: No @cdeshmukh , the requirement from the bug is , if the service remains in say in activating state for more that 180 secs, then we used to give fault description as `"statsd.service in activating state for more than 180 seconds"`. But with the latest changes by Sandeep, by mistake, we used the `FaultAlert` msg template instead of `InactiveAlert` template, so it gives `"statsd.service in activating state."` as a description.
So we just needed to use that `InactiveAlert` description to correct it.

```Before raising fault-resolved alert, logic should wait n see if service stays recovered for sometime and add in description about this time wait too.```
This is the new thing that we can think of addding.

Comment (shriya-deshmukh) [reply low-level/sensors/impl/centos_7/service_monitor.py:449]: With current implementation:
We wait fo threshhold time(180 sec) only if service is in intermediate state i.e - inactive, deactivating, activating.
As soon we detect service state changed from any intermediate state to active or failed to active, we raised the fault_resolved alert.
Should I add wait for fault_resolved case also??
@cdeshmukh

Comment (cdeshmukh) [reply low-level/sensors/impl/centos_7/service_monitor.py:449]: >But with the latest changes by Sandeep, by mistake, we used the FaultAlert msg template instead of InactiveAlert template, so it gives "statsd.service in activating state." as a description.
So we just needed to use that InactiveAlert description to correct it.

@sumedhak27 of then lets correct that, but bug is about fault-resolved mainly

>Should I add wait for fault_resolved case also??

@shriya-deshmukh  yes, thats what i meant. to avoid raising too many alerts if state transition happens quite often. rare but still possible

**Review by jjxsg — APPROVED** [review]

File: low-level/sensors/impl/centos_7/service_monitor.py (hunk 1 above)
Comment (shriya-deshmukh) [reply low-level/sensors/impl/centos_7/service_monitor.py:462]: Done, added wait/threshold time for fault-resolved alert and updated description as.
ie. 
ResolvedAlert = Alert("fault_resolved", **"{} in {} state from last {} seconds."**,
"{} service is available now.",
"No action required")

File: low-level/sensors/impl/centos_7/service_monitor.py [hunk 2]
```diff
@@ -169,22 +173,29 @@ def __init__(self, unit):
...
+                     f"{ServiceMonitor.name().upper()}>threshold_active_time",
+                     '30'))
         self.properties_changed_signal = None
         self._service_state = ActiveState
         self._unit_state = None
 
     @property
     def is_enabled(self):
-        return 'disabled' not in self.properties_iface.Get(
+        return 'enabled' in self.properties_iface.Get(
```
shriya-deshmukh, line 190: Modified condition here as - 
self.properties_iface.Get(UNIT_IFACE, 'UnitFileState') returns nothing if service UnitFile is not present for service (Services which are not configured )
For such services previous condition return TRUE and we are getting alerts, event if service is not present on setup.

cdeshmukh: If a service, which needs to be monitored, is not configured, it will be not able recognized i believe. How/why this condition should return 'enabled' then in such case ?

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -216,9 +227,8 @@ def properties_changed_handler(self, interface, changed_properties,
 
         if self.state != self.previous_state:
             if self.state == "active":
-                if self._service_state is FailedState:
-                    Service.alerts.put(
-                        ServiceMonitor.get_alert(self, ResolvedAlert))
+                if self._service_state is FailedState and self.name not in Service.active_services:
+                    Service.active_services.add(self.name)
```
cdeshmukh, line 231: why service added in active list, when its state is failed?
Also raising alert is avoided here, Line 221 ?

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -446,7 +459,20 @@ def check_nonactive_services(self):
...
+        """
+            Monitor active services.
+
+            Raise fault resolved alert for active services
+            if service stays in active state for threshhold time.
+        """
+        for service in Service.active_services.copy():
+            if self.services[service].is_active_for_threshold_time():
+                self.raise_alert(
+                    self.get_alert(self.services[service], ResolvedAlert))
```
cdeshmukh, line 474: Ok alert raising code is shifted here

low-level/sensors/impl/centos_7/service_monitor.py
shriya-deshmukh: **'disabled' not in self.properties_iface.Get(UNIT_IFACE, 'UnitFileState')** 
Services which are present/configured on system only for those services this condition works.

for eg: 
statsd.service is configured and in enabled/disabled state, then this condition will work as expected.
But if 'lnet.service' is not configured on system/ (service unit is not present) in that case,
'properties_iface.Get(UNIT_IFACE, 'UnitFileState')' this return nothing, and 
'disbaled not in properties_iface.Get(UNIT_IFACE, 'UnitFileState') ' this condition will be true, 
and we get the alert for service even that is not configured on system.

```
>>> service_name=lnet.service
>>> str(properties_iface.Get(UNIT_IFACE, 'UnitFileState'))
''
>>> 'disabled' not in properties_iface.Get(UNIT_IFACE, 'UnitFileState')
True
>>> 'enabled' in properties_iface.Get(UNIT_IFACE, 'UnitFileState')
False
>>>
```

As per my understanding from monitoring_service list , services which are in enabled state only those services need to be monitored.
If any service from monitoring_service list is not configured on system, do we need to raise fault alert for such services??
if yes then in that case previous('disabled' not in self.properties_iface.) condition will work.

low-level/sensors/impl/centos_7/service_monitor.py
shriya-deshmukh: `why service added in active list, when its state is failed?`

We are adding service in active_services set if current state of service is 'active'. Line-229.

`Also raising alert is avoided here, Line 221 ?`
this has been moved to **check_active_services** function. Line-464

low-level/sensors/impl/centos_7/service_monitor.py
jjxsg: [UnitFileState](https://www.freedesktop.org/wiki/Software/systemd/dbus/) can have these values: 
enabled, enabled-runtime, linked, linked-runtime, masked, masked-runtime, static, disabled

If except for disabled, a service will be classified as enabled for all other states. So,
```
'enabled' in self.properties_iface.Get(UNIT_IFACE, 'UnitFileState')
```
may not always give right results. 

In node health, I put additional check, if `command_line_path` exists for a service or not. If a service is not configured it wont exist.

```
path_array = properties_iface.Get(const.SERVICE_IFACE, 'ExecStart')
try:
    command_line_path = str(path_array[0][0])
except IndexError as err:
    logger.error(self.log.svc_log(f"Unable to find {service_name} path due to {err}"))
    command_line_path = "NA"
```

shriya-deshmukh: Okay Thanks @jjxsg , I will add same comand_line_path check here to check if service is configured or not.
```
try:
            command_line_path = str(path_array[0][0])
        except IndexError as err:
            logger.error(
                "Unable to find %s path due to %s" % (self.name, err))
            command_line_path = "NA"
        if command_line_path != "NA":
            return 'disabled' not in self.properties_iface.Get(
                UNIT_IFACE, 'UnitFileState')
        else:
            return False
```

sumedhak27: Hi @shriya-deshmukh , @jjxsg  I am bit confused here, we need review from @sandeepsgt about this.

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -461,6 +499,12 @@ def raise_iem(self, service, alert_type):
 
     @classmethod
     def get_alert(cls, service, alert):
+        if service.state == "active":
+            description = alert.description.format(
+                service.name, service.state, service.active_threshold)
+        else:
+            description = alert.description.format(
+                service.name, service.state, service.nonactive_threshold)
```
sumedhak27, line 507: Hi, what happens in case of `FailedAlert`? it has only two format args to replace.
`"{} is in {} state"`

jjxsg (approved):

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -169,22 +173,41 @@ def __init__(self, unit):
...
+        try:
+            command_line_path = str(path_array[0][0])
+        except IndexError as err:
+            logger.error(
+                "Unable to find %s path due to %s" % (self.name, err))
+            command_line_path = "NA"
+        if command_line_path != "NA":
+            return 'disabled' not in self.properties_iface.Get(
+                UNIT_IFACE, 'UnitFileState')
+        else:
```
sumedhak27, line 202: `return False` can be outside of else as well.

sumedhak27 (approved):

low-level/sensors/impl/centos_7/service_monitor.py
sandeepsgt: @shriya-deshmukh  what if you add check like below

if self.properties_iface.Get(UNIT_IFACE, 'UnitFileState'):
    return 'disabled' not in self.properties_iface.Get(.....
else:
    return False

shriya-deshmukh: Yes, It works. 
Added code change.

low-level/sensors/impl/centos_7/service_monitor.py
shriya-deshmukh: done

low-level/sensors/impl/centos_7/service_monitor.py
cdeshmukh: >We are adding service in active_services set if current state of service is 'active'. Line-229.
I dont understand this still, line 229 must have changed with code changes, no code there now.

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -169,22 +173,31 @@ def __init__(self, unit):
...
+                     '30'))
         self.properties_changed_signal = None
         self._service_state = ActiveState
         self._unit_state = None
 
     @property
     def is_enabled(self):
-        return 'disabled' not in self.properties_iface.Get(
-            UNIT_IFACE, 'UnitFileState')
+        if self.properties_iface.Get(UNIT_IFACE, 'UnitFileState'):
```
cdeshmukh, line 190: Did @sandeepsgt confirmed if this is fine. Also i had some comment around this i believe

shriya-deshmukh: Yes, He himself only suggested above change. [comment](https://github.com/Seagate/cortx-monitor/pull/444#discussion_r666803104) @sandeepsgt  please confirm.
I have tested it, with tis change we raise alert for services which are configured on system.

low-level/sensors/impl/centos_7/service_monitor.py
shriya-deshmukh: Okay let me explain with example.
If service detected Failed/Inactive for threshold time, service state changed to FailedState (self._service_state=FailedState). 
Now when in next iteration service state change detected, 
here in properties_changed_handler function we fetch current state,substate pid of service.
then check if current state not equals to prev_state, and if current service state is active then check
 self.service_state = FailedState and service name not active service list, then only we add it into active_service set.
 and in next line chnged service state to self.service_state = ActiveState. 
 
```
 if self.state != self.previous_state:
            if self.state == "active":
                if self._service_state is FailedState and self.name not in Service.active_services:
                    Service.active_services.add(self.name)
                self.new_service_state(ActiveState)
```

sandeepsgt (approved):

jjxsg (approved):

low-level/sensors/impl/centos_7/service_monitor.py
cdeshmukh: Got now after our discussion @shriya-deshmukh 
Only if service file is present, then check if disabled in that service file & return true/false accordingly , else return false.

low-level/sensors/impl/centos_7/service_monitor.py
cdeshmukh: Understand now, just one comment. Better to do
>self.new_service_state(ActiveState)

first, before
> self.new_service_state(ActiveState)

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -216,10 +229,11 @@ def properties_changed_handler(self, interface, changed_properties,
...
             if self.state == "active":
-                if self._service_state is FailedState:
-                    Service.alerts.put(
-                        ServiceMonitor.get_alert(self, ResolvedAlert))
-                self.new_service_state(ActiveState)
+                if self._service_state is FailedState and self.name not in Service.active_services:
+                    self.new_service_state(ActiveState)
+                    Service.active_services.add(self.name)
+                if self._service_state is InactiveState:
+                    self.new_service_state(ActiveState)
```
cdeshmukh, line 236: > Service.active_services.add(self.name)

Adding to active service needed here also i believe @line 232
>  if self._service_state in (FailedState, InactiveState) and self.name not in Service.active_services:

In which case, can easily club the condition i believe

low-level/sensors/impl/centos_7/service_monitor.py
sandeepsgt: `Service.active_services.add(self.name)`

Can't you  add this into ActiveService class's enter method ?

low-level/sensors/impl/centos_7/service_monitor.py
shriya-deshmukh: No, Here don't need to add service in active_service list, because service will be in InactiveState only till threshold_time after that service_state change to FailedState, then we raise fault alert.

But if service state change from inactive to active before threshold time, then we will not raise any fault alert, and hence fault_resolved alert also not needed, thats why service not added in active_service list.

cdeshmukh: oh right, agree. Guess we are good.

low-level/sensors/impl/centos_7/service_monitor.py
shriya-deshmukh: Service get added into list only if there is previous fault_alert detected for service and current state is active, only for such service we raise fault_resolved alert.
If I move this to ActiveService class's enter method then all active services gets added into this list and we will get fault_resolved alert for all this services even if there is no fault alert detected.

cdeshmukh (approved):

cdeshmukh:
test this please

sandeepsgt (approved):

PR #448, opened 2021-07-07
EOS-22214: Fix adaptive config

sandeepsgt:
Jira :[ EOS-22214](https://jts.seagate.com/browse/EOS-22214)
## Problem Statement:
global_config_copy_url was not present in default config file. So, conf_upgrade script treated that as removed key and hence it was not added into existing config file. SSPL service failed because of this

## Solution Design:

Add global_config_copy_url in default config file with "" value

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?

### DIFF
--- a/low-level/files/opt/seagate/sspl/conf/sspl.conf.LR2.yaml
+++ b/low-level/files/opt/seagate/sspl/conf/sspl.conf.LR2.yaml
@@ -192,6 +192,7 @@ SYSTEM_INFORMATION:
    environment: PROD
    sysfs_base_path: /sys/
    sspl_state: active
+   global_config_copy_url: ""
 
 # NOTE: Keys present in OBSOLETE section are out of date
 # and will be removed in furture version. Keys present in


low-level/files/opt/seagate/sspl/conf/sspl.conf.LR2.yaml
```diff
@@ -192,6 +192,7 @@ SYSTEM_INFORMATION:
    environment: PROD
    sysfs_base_path: /sys/
    sspl_state: active
+   global_config_copy_url: ""
```
cdeshmukh, line 195: why copy url path needed in config. If not seen in config, code can handle it right ?

sandeepsgt: if global_config_copy_url(any key)  is not in default config file but it is in existing config. conf_upgrade script treat that as removed key. So, it will  not get added into existing config file. SSPL service failed because of this.
global_config_copy_url key get added later in provisioning steps.

sumedhak27 (approved):

low-level/files/opt/seagate/sspl/conf/sspl.conf.LR2.yaml
cdeshmukh: Ok so the value gets  updated during provisioning, then looks fine.

cdeshmukh (approved):

thavanathan (approved):

PR #449, opened 2021-07-07
EOS-21004: Introduce waiting time for CPU and memory usage fault resolved alert

jjxsg:
## Problem Statement:

Fault resolved alert for CPU/memory usage should wait for certain waiting threshold time.

## Solution Design:

Introduced waiting time for CPU/memory usage fault resolved alerts.

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?

### DIFF
--- a/low-level/message_handlers/node_data_msg_handler.py
+++ b/low-level/message_handlers/node_data_msg_handler.py
@@ -458,7 +458,8 @@ class NodeDataMsgHandler(ScheduledModuleThread, InternalMsgQ):
         else:
             fault_resolved_iters = 0
         try:
-            iteration_limit = int(self._high_memory_usage_wait_threshold/self._transmit_interval)
+            iteration_limit = int(
+                self._high_memory_usage_wait_threshold/self._transmit_interval)
         except ZeroDivisionError:
             iteration_limit = 0
         self.usage_time_map['memory'] = current_time
--- a/low-level/message_handlers/node_data_msg_handler.py
+++ b/low-level/message_handlers/node_data_msg_handler.py
@@ -87,6 +87,11 @@ class NodeDataMsgHandler(ScheduledModuleThread, InternalMsgQ):
         'memory' : -1,
         'disk' : -1
     }
+    fault_resolved_iterations = {
+        'cpu': 0,
+        'memory': 0,
+        'disk': 0
+    }
     prev_nw_status = {}
     prev_cable_cnxns = {}
     # Dir to maintain fault detected state for interface
@@ -132,7 +137,7 @@ class NodeDataMsgHandler(ScheduledModuleThread, InternalMsgQ):
         super(NodeDataMsgHandler, self).initialize_msgQ(msgQlist)
 
         self._transmit_interval = int(Conf.get(SSPL_CONF, f"{self.NODEDATAMSGHANDLER}>{self.TRANSMIT_INTERVAL}",
-                                                60))
+                                               60))
         self._high_cpu_usage_wait_threshold = int(Conf.get(SSPL_CONF,
                                                 f"{self.NODEDATAMSGHANDLER}>{self.HIGH_CPU_USAGE_WAIT_THRESHOLD}",60))
         self._high_memory_usage_wait_threshold = int(Conf.get(SSPL_CONF,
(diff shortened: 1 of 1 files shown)


sandeepsgt (approved):
LGTM

low-level/message_handlers/node_data_msg_handler.py
```diff
@@ -443,6 +451,16 @@ def _generate_host_update(self):
...
         else:
             previous_check_time = int(-1)
+        if memory_persistent_data\
+                .get('memory_fault_resolved_iterations') is not None:
+            fault_resolved_iters = int(
+                memory_persistent_data['memory_fault_resolved_iterations'])
+        else:
+            fault_resolved_iters = 0
+        try:
+            iteration_limit = int(self._high_memory_usage_wait_threshold/self._transmit_interval)
```
sumedhak27, line 461: Hi, @jjxsg , personally I feel following way easier, than calculating iteration,
```python
if time.time() - fault_resolved_time > wait_threshold:
    raise alert
```
But I don't know which way is better?

jjxsg: Time based solution is what I used for fault alerts which is still there. In that case also we need to persist the fault resolved time in case fault comes again within the threshold waiting time. 
Iteration based approach handles the case when system or SSPL restarts within the waiting time interval for fault resolved alert. 

Let's say x is the time when fault is resolved and waiting time is 60 seconds which means fault resolved alert should be sent at time x+60.  Let's say reboot occurred at x+20 which took another 2 mins to restore the service in previous state. 
Now while checking whether to wait or send the alert, it will compare x+60 with (x+20+120) which is obviously greater, so it will send the alert immediately. But expected behavior is that it should wait for another 40 seconds before sending the alert.

sumedhak27:
```
it will compare x+60 with (x+20+120) which is obviously greater, so it will send the alert immediately. 
But expected behavior is that it should wait for another 40 seconds before sending the alert.
```
Ohh really!! I was unaware of this expected behavior, I thought the service should be up for say 60s irrespective of whether sspl is observing it or not. 
Ideally, I think whether sspl is watching the service or not should not matter, we can directly check the uptime of the service and if the service is up for more than 60 sec then we can raise fault_resolved. This way we won't have to save time or count to cache as well.

jjxsg: Well, this is true for a service because its state doesn't fluctuate frequently. But CPU and Memory usage fluctuate, so we need to check the usage in real time rather than depending on past values.

sumedhak27: Ohh sorry this is related to cpu and memory usage, my bad.😅 Then that makes sense, Thanks the patient explanation @jjxsg 👍

sumedhak27 (approved):

shriya-deshmukh (approved):

low-level/message_handlers/node_data_msg_handler.py
```diff
@@ -494,41 +513,47 @@ def _generate_host_update(self):
...
+                    self.persist_state_data('memory', 'MEMORY_USAGE_DATA')
+                elif fault_resolved_iters >= iteration_limit:
+                    # Create the memory data message and hand it over
+                    # to the egress processor to transmit
+                    fault_resolved_event = "Host memory usage has decreased to {}%, "\
+                        "lower than the configured threshold of {}%.".format(
+                            self._node_sensor.total_memory["percent"],
+                            self._host_memory_usage_threshold
+                        )
+                    logger.warning(fault_resolved_event)
```
cdeshmukh, line 528: Why warning, its a good event, can be logger.info !
I see this is how it was from earlier

low-level/message_handlers/node_data_msg_handler.py
```diff
@@ -641,46 +676,53 @@ def _generate_cpu_data(self):
...
+                elif fault_resolved_iters >= iteration_limit:
+
+                    # Create the cpu usage data message and hand it over
+                    # to the egress processor to transmit
+                    fault_resolved_event = "CPU usage has decreased to {}%, "\
+                        "lower than the configured threshold of {}%.".format(
+                            self._node_sensor.cpu_usage,
+                            self._cpu_usage_threshold
+                        )
+                    logger.warning(fault_resolved_event)
```
cdeshmukh, line 692: logger.info

cdeshmukh (approved):

low-level/message_handlers/node_data_msg_handler.py
jjxsg: Done.

low-level/message_handlers/node_data_msg_handler.py
jjxsg: Done

low-level/message_handlers/node_data_msg_handler.py
jjxsg: Updated for disk fault resolved event as well.

cdeshmukh:
test this please

thavanathan (approved):

PR #453, opened 2021-07-14
EOS-21171:Node down IEM/Alert is lost in SSPL

shriya-deshmukh:
## Problem Statement:

Instance creation for MessageProducer class failed when kafka is down on all three nodes.

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?

### DIFF
--- a/low-level/framework/messaging/egress_accumulated_msgs_processor.py
+++ b/low-level/framework/messaging/egress_accumulated_msgs_processor.py
@@ -131,6 +131,7 @@ class EgressAccumulatedMsgsProcessor(ScheduledModuleThread, InternalMsgQ):
                     if isinstance(self._producer, MessageProducer):
                         self._producer.send([message])
                         logger.error(f"Published Accumulated Message {message}")
+                        self.store_queue.delete()
                     else:
                         self.create_MsgProducer_obj()
         except MessageBusError as e:
--- a/low-level/framework/messaging/egress_accumulated_msgs_processor.py
+++ b/low-level/framework/messaging/egress_accumulated_msgs_processor.py
@@ -74,9 +74,15 @@ class EgressAccumulatedMsgsProcessor(ScheduledModuleThread, InternalMsgQ):
         self.store_queue = StoreQueue()
         self._read_config()
         producer_initialized.wait()
-        self._producer = MessageProducer(producer_id="acuumulated processor",
-                                         message_type=self._message_type,
-                                         method=self._method)
+        self.create_MsgProducer_obj()
+
+    def create_MsgProducer_obj(self):
+        self._producer = None
+        try:
+            self._producer = MessageProducer(producer_id=self._producer_id,
+                message_type=self._message_type, method=self._method)
+        except Exception as err:
+            logger.error('Instance creation for MessageProducer class failed due to %s' % err)
 
     def read_data(self):
         """This method is part of interface. Currently it is not
@@ -122,7 +128,11 @@ class EgressAccumulatedMsgsProcessor(ScheduledModuleThread, InternalMsgQ):
                             continue
                     if "sensor_response_type" in dict_msg["message"]:
                         logger.info(f"Publishing Accumulated Alert: {message}")
-                    self._producer.send([message])
+                    if isinstance(self._producer, MessageProducer):
+                        self._producer.send([message])
+                        logger.error(f"Published Accumulated Message {message}")
+                    else:
+                        self.create_MsgProducer_obj()
         except MessageBusError as e:
             logger.error("EgressAccumulatedMsgsProcessor, run, %r" % e)
         except Exception as e:
--- a/low-level/framework/messaging/egress_accumulated_msgs_processor.py
+++ b/low-level/framework/messaging/egress_accumulated_msgs_processor.py
@@ -130,7 +130,7 @@ class EgressAccumulatedMsgsProcessor(ScheduledModuleThread, InternalMsgQ):
                         logger.info(f"Publishing Accumulated Alert: {message}")
                     if isinstance(self._producer, MessageProducer):
                         self._producer.send([message])
-                        logger.error(f"Published Accumulated Message {message}")
+                        logger.info(f"Published Accumulated Message {message}")
                         self.store_queue.delete()
                     else:
                         self.create_MsgProducer_obj()
--- a/low-level/framework/messaging/egress_processor.py
+++ b/low-level/framework/messaging/egress_processor.py
@@ -81,10 +81,16 @@ class EgressProcessor(ScheduledModuleThread, InternalMsgQ):
         self._request_shutdown = False
 
         self._read_config()
-        self._producer = MessageProducer(producer_id=self._producer_id,
-                                         message_type=self._message_type,
-                                         method=self._method)
+        self.create_MsgProducer_obj()
         producer_initialized.set()
+
+    def create_MsgProducer_obj(self):
+        self._producer = None
+        try:
+            self._producer = MessageProducer(producer_id=self._producer_id,
+                message_type=self._message_type, method=self._method)
+        except Exception as err:
+            logger.error('Instance creation for MessageProducer class failed due to %s' % err)
 
     def run(self):
         """Run the module periodically on its own thread. """
@@ -205,7 +211,10 @@ class EgressProcessor(ScheduledModuleThread, InternalMsgQ):
                          "actuator_response_type").get(
                          "thread_controller") is not None):
                 self._add_signature()
-                self._producer.send([json.dumps(self._jsonMsg)])
+                if isinstance(self._producer, MessageProducer):
+                    self._producer.send([json.dumps(self._jsonMsg)])
+                else:
+                    self.create_MsgProducer_obj()
                 logger.debug(
                     "_transmit_msg_on_exchange, Successfully Sent: %s" % self._jsonMsg)
             else:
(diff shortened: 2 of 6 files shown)


low-level/framework/messaging/egress_accumulated_msgs_processor.py
```diff
@@ -122,7 +128,11 @@ def run(self):
                             continue
                     if "sensor_response_type" in dict_msg["message"]:
                         logger.info(f"Publishing Accumulated Alert: {message}")
-                    self._producer.send([message])
+                    if isinstance(self._producer, MessageProducer):
+                        self._producer.send([message])
```
sandeepsgt, line 132: If this fails, what happens to current message being processed ?

low-level/framework/messaging/egress_processor.py
```diff
@@ -205,16 +211,22 @@ def _transmit_msg_on_exchange(self):
...
                 jsonMsg = json.dumps(self._jsonMsg)
                 try:
                     if self.store_queue.is_empty():
-                        self._producer.send([jsonMsg])
-                        logger.info(f"Published Alert: {jsonMsg}")
+                        if isinstance(self._producer, MessageProducer):
+                            self._producer.send([jsonMsg])
+                            logger.info(f"Published Alert: {jsonMsg}")
+                        else:
+                            self.create_MsgProducer_obj()
```
sandeepsgt, line 229: Same quesiton, If this fails or pass, what happens to current message being processed ?

low-level/framework/messaging/egress_accumulated_msgs_processor.py
shriya-deshmukh: As discussed, changed get method in store_queue, and added separate delete method for deleting an element from the queue
Deleting element only if msg successfully sent - line134.

low-level/framework/messaging/egress_processor.py
shriya-deshmukh: if msg_producer instance is not available then in 'else' part, adding that msg in accumulated queue, sit will be processed by accumulated processor.

low-level/framework/messaging/egress_accumulated_msgs_processor.py
```diff
@@ -122,7 +128,12 @@ def run(self):
...
+                        logger.error(f"Published Accumulated Message {message}")
```
cdeshmukh, line 133: line no. 130 log not enough? also why log.error?

low-level/framework/messaging/egress_accumulated_msgs_processor.py
```diff
@@ -122,7 +128,12 @@ def run(self):
...
+                        self.store_queue.delete()
+                    else:
+                        self.create_MsgProducer_obj()
```
cdeshmukh, line 136: if code reaches here, then current message wont be procedded rt !? until all existing queue messages are iterated through

low-level/framework/messaging/egress_accumulated_msgs_processor.py
```diff
@@ -74,9 +74,15 @@ def initialize(self, conf_reader, msgQlist, products):
...
         producer_initialized.wait()
-        self._producer = MessageProducer(producer_id="acuumulated processor",
-                                         message_type=self._message_type,
-                                         method=self._method)
+        self.create_MsgProducer_obj()
+
+    def create_MsgProducer_obj(self):
+        self._producer = None
+        try:
+            self._producer = MessageProducer(producer_id=self._producer_id,
```
cdeshmukh, line 82: self._producer_id
where this is declared

low-level/framework/messaging/egress_accumulated_msgs_processor.py
shriya-deshmukh: Corrected log level, changed it to logger.info.
>>line no. 130 log not enough? 
No this I have added to make sure producer.send() executed successfully.
So if we see this log in log file, then we can say from sspl side we have sent an msg successfully.

low-level/framework/messaging/egress_accumulated_msgs_processor.py
shriya-deshmukh: yes, right.

cdeshmukh: and picked up again in next iteration. so order of message delivery from accumulated queue is disturbed. we need to maintain the order

low-level/framework/messaging/egress_processor.py
```diff
@@ -81,11 +81,17 @@ def initialize(self, conf_reader, msgQlist, product):
         self._request_shutdown = False
 
         self._read_config()
-        self._producer = MessageProducer(producer_id=self._producer_id,
-                                         message_type=self._message_type,
-                                         method=self._method)
+        self.create_MsgProducer_obj()
```
cdeshmukh, line 84: can we have one function somewhere, both for consumer & producer, which can used from all places where ever needed

low-level/framework/messaging/egress_accumulated_msgs_processor.py
```diff
@@ -122,7 +128,12 @@ def run(self):
                             continue
                     if "sensor_response_type" in dict_msg["message"]:
                         logger.info(f"Publishing Accumulated Alert: {message}")
-                    self._producer.send([message])
+                    if isinstance(self._producer, MessageProducer):
+                        self._producer.send([message])
```
cdeshmukh, line 132: create_MsgProducer_obj() here and then try send

low-level/framework/messaging/egress_accumulated_msgs_processor.py
shriya-deshmukh: Inside _read_config method.
line 157.

low-level/framework/messaging/egress_accumulated_msgs_processor.py
shriya-deshmukh: Why here??
So 1st time we are calling 'create_MsgProducer_obj()' in initialize() method, and 
here we are checking if self._producer instance is available, if it is not available then we are calling 'create_MsgProducer_obj()' function in else.

low-level/framework/messaging/egress_accumulated_msgs_processor.py
shriya-deshmukh: message delivery from accumulated queue will not disturbed,  because we delete the msg only in case of successful send.
eg: if there are 3 msgs in queue m1,m2,m3,
 first we read msg from queue - m1, if it is not send successfully in 2nd iteration also we will read m1 only.
 and whatever new alert msgs are detected by sspl that will get added into accumulated queue because we  in egress_processor have check that if queue is not empty msg will be added into queue, 
 this way we are maintaining order.
 
 @sandeepsgt please correct me if I am wrong.

low-level/framework/messaging/egress_processor.py
shriya-deshmukh: Yes, we can have, where we can define this??
should I create separate class inside - low-level/framework/messaging/create_msgbus_instance.py
`class MessageBusInstance:
      def create_MsgProducer_obj():
      
      def create_MsgConsumer_obj():
`

low-level/framework/messaging/egress_accumulated_msgs_processor.py
sandeepsgt: Correct

low-level/framework/messaging/egress_accumulated_msgs_processor.py
sandeepsgt: Ok. This looks fine

sandeepsgt (approved):

low-level/framework/messaging/egress_accumulated_msgs_processor.py
cdeshmukh: got it

low-level/framework/messaging/egress_accumulated_msgs_processor.py
cdeshmukh: As i understood in discussion, until delete in invoked, message from queue wont disappear and same message will be picked up as next element in same iteration also. in which case, alert order processing is not affected

low-level/framework/messaging/egress_accumulated_msgs_processor.py
cdeshmukh: understood. as discussed, am assuming this is tested by now

low-level/framework/messaging/egress_processor.py
cdeshmukh: as discussed, no. New class just to maintain a function to create an object, doesnt make sense. So let it as it is today

cdeshmukh (approved):

low-level/framework/messaging/egress_accumulated_msgs_processor.py
shriya-deshmukh: yes.

low-level/framework/messaging/ingress_processor.py
```diff
@@ -108,10 +108,17 @@ def initialize(self, conf_reader, msgQlist, product):
...
+
+    def create_MsgConsumer_obj(self):
+        self._consumer = None
+        try:
+            self._consumer = MessageConsumer(consumer_id=self._consumer_id,
+                consumer_group=self._consumer_group,
+                message_types=[self._message_type],
+                auto_ack=False, offset=self._offset)
+        except Exception as err:
+            logger.error('Instance creation for MessageConsumer class failed due to %s' % err)
```
jjxsg, line 121: Ideally, we should avoid generic exceptions. Please check if you can put specific exception here, otherwise good to go.

jjxsg (approved):

low-level/framework/messaging/ingress_processor.py
shriya-deshmukh: Here we got error- 'KafkaError{code=_ALL_BROKERS_DOWN,val=-187,str="1/1 brokers are down"}' which is not getting handled by MessageBusError exception, hence added generic exception.

jjxsg: Okay.

cdeshmukh:
VM deployment was seen fine before PR was rebased, going ahead with the merge

PR #468, opened 2021-07-28
EOS-23158: Alerts plan tests(CPU, Memory and IEM)  are failing on VM

mariyappanp:
## Problem Statement:

EOS-23158: Alerts plan tests(CPU, Memory and IEM)  are failing on VM

## Coding:

## Testing:

## Integration:

* [x] If there is any interface change, did you communicate it to other components?
* [x] If there is any interface change, other componenent gate gatekeepers are ready to accept the change?
* [x] Did you complete deployment test if applicable?

## PR checklist:

* [x] DCO and cla-signed?


jjxsg (approved):

gargpalak048 (approved):

low-level/sensors/impl/generic/node_data.py
```diff
@@ -157,11 +157,14 @@ def _get_host_update_data(self):
...
-                                       psutil.STATUS_SLEEPING):
-                total_running_proc += 1
+            try:
+                pinfo = proc.as_dict(attrs=['status'])
+                if pinfo['status'] not in (psutil.STATUS_ZOMBIE, psutil.STATUS_DEAD,
+                                        psutil.STATUS_STOPPED, psutil.STATUS_IDLE,
+                                        psutil.STATUS_SLEEPING):
+                    total_running_proc += 1
+            except psutil.NoSuchProcess:
+                logger.warn(f"(psutil) Process '{proc.name()}' exited unexpectedly.")
```
sumedhak27, line 167: Hi, I think we should also have generic exception handling in case to avoid failure.

sumedhak27 (approved):

shriya-deshmukh (approved):

low-level/sensors/impl/generic/node_data.py
mariyappanp: No such situation occurred so far. 
But this NoSuchProcess is identified during sspl_test restart sspl service and ERROR is logged. This is expected so we need to handle it.
In case of generic exception, it can occur anywhere in SSPL and its already handled in sspl_ll_d.

jjxsg: @sumedhak27 Its a rare error in case a process gets exited while checking running process.
`psutil.process_iter()` is not robust against exiting process, so it needs to be in try/exception and our goal is to minimize the use of generic exceptions, that's why exception supplied from the library is used. 
Ref. 
https://github.com/giampaolo/psutil/issues/1811

sumedhak27: Hi, recently in iem case, we encountered the same issue, since we were handling only the specific exception, for an unhandled specific exception, which is caught in sspl_ll_d the whole sensor was turned off, so finally, Shriya added the generic expectation along with specific one, I wanted to mention such case @mariyappanp.

mariyappanp: Okay, Sumedh.

jjxsg:
@mariyappanp currently sanity tests do not give any info why a test got failed and with too many logs, it becomes extremely hard to debug. We can add something like `traceback.print_exc()` within test exception to print the actual cause of failure. Please check if you can include this improvement in current PR.

jjxsg (approved):

sspl_test/run_test.py
```diff
@@ -92,6 +92,7 @@ def tmain(argp, argv):
             init(args)
         except Exception as e:
             print('FAILED: Error: %s #@#@#@' %e)
+            print(traceback.format_exc())
```
cdeshmukh, line 95: This will traceback for failed test only right, and not impacthe final test results list shown at last? In which case am ok

mariyappanp: Yes, only for the failed test. Final test result will come as usual.

PR #489, opened 2021-08-05
EOS-23182: Improve Alerts Description - Server FRU Alerts - Fans

gargpalak048:
## Problem Statement:

Improve Alerts Description - Server FRU Alerts - Fans

## Solution Design:

Updated description & added 'impact' and 'recommendation' field to alert schema

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?

### DIFF
--- a/low-level/sensors/impl/generic/node_hw.py
+++ b/low-level/sensors/impl/generic/node_hw.py
@@ -959,19 +959,21 @@ class NodeHWsensor(SensorThread, InternalMsgQ):
         original_event = event
         original_status = status
 
-        if status.lower() == "deasserted" and event.lower() == "fully redundant":
-            alert_type = "fault"
-        elif status.lower() == "asserted" and event.lower() == "fully redundant":
-            alert_type = "fault_resolved"
-        elif "going high" in event.lower() or "going low" in event.lower():
-            if status.lower() == "deasserted":
-                alert_type = "fault_resolved"
-                event = "Deasserted - " + event
-            elif status.lower() == "asserted":
-                alert_type = "fault"
-                event = "Asserted - " + event
-
         sensor_name = self.sensor_id_map[self.TYPE_FAN][sensor_id]
+        fan_state=None
+        for specific in fan_specific_list:
+            if specific.lower() in event.lower():
+                fan_state = specific
+                break
+        try:
+            alert = alert_for_event[self.TYPE_FAN][fan_state][status]
+        except KeyError:
+            logger.error(f"{self.TYPE_FAN} : " +
+                         f"Unknown event: {fan_state}, status: {status}")
+            return
+        except Exception as e:
+            logger.exception(e)
+            return
 
         fan_common_data, fan_specific_data, fan_specific_data_dynamic = self._get_sensor_props(sensor_name)
 
@@ -980,40 +982,33 @@ class NodeHWsensor(SensorThread, InternalMsgQ):
                 del fan_specific_data[key]
 
         fan_info = fan_specific_data
-        fan_info.update({"fru_id" : device_id, "event" : event})
+        fan_info.update({"fru_id" : device_id, "event" : status + " - " + event})
         resource_type = NodeDataMsgHandler.IPMI_RESOURCE_TYPE_FAN
         fru = self.ipmi_client.is_fru(self.fru_map[self.TYPE_FAN])
-        severity_reader = SeverityReader()
-        if alert_type:
-            severity = severity_reader.map_severity(alert_type)
-        else:
-            # Else section will handle some unknown as well as known events like
-            # "State Asserted", "State Deasserted" or "Sensor Reading" and also it is
-            # keeping default severity and alert_type for those events.
-            alert_type = "miscellaneous"
-            severity = "informational"
 
         fru_info = {    "resource_type": resource_type,
                         "fru": fru,
                         "resource_id": sensor_name,
                         "event_time": self._get_epoch_time_from_date_and_time(date, _time),
-                        "description": event
+                        "description": alert.description.format(sensor_name, device_id),
+                        "impact": alert.impact,
+                        "recommendation": alert.recommendation
                     }
 
         if is_last:
             fan_info.update(fan_specific_data_dynamic)
 
-        if alert_type == "fault":
+        if alert.type == "fault":
             self.faulty_resources[sensor_name] = {
                 'status': original_status,
                 'event': original_event,
                 'device_id': device_id,
                 'fru_type': self.TYPE_FAN
             }
-        elif alert_type in ['fault_resolved', 'miscellaneous'] and sensor_name in self.faulty_resources:
+        elif alert.type in ['fault_resolved'] and sensor_name in self.faulty_resources:
             del self.faulty_resources[sensor_name]
 
-        self._send_json_msg(resource_type, alert_type, severity, fru_info, fan_info)
+        self._send_json_msg(resource_type, alert.type, alert.severity, fru_info, fan_info)
         store.put(self.faulty_resources, self.faulty_resources_path)
 
     def _parse_psu_supply_info(self, index, date, _time, sensor, sensor_num, event, status, is_last):
--- a/low-level/sensors/impl/generic/node_hw_alerts_info.py
+++ b/low-level/sensors/impl/generic/node_hw_alerts_info.py
@@ -135,5 +135,114 @@ alert_for_event = {
                 "Power Supply regained the healthy state.", "None"
             )
         })
+    },
+    "Fan": {
+        **dict.fromkeys(["Fully Redundant", "Fully Redundant ()"], {
+            "Asserted": Alert(
+                "fault_resolved", "informational",
+                "'{} Sensor' has recovered from a failure. [{}]",
+                "None", "None"),
+            "Deasserted": Alert(
+                "fault", "critical",
+                "'{} Sensor', is Fully redundant . [{}]",
+                "Fan efficiency will be impacted",
+                DEFAULT_RECOMMENDATION)
+        }),
+        **dict.fromkeys(["Lower Non-Recoverable", "Lower Non-Recoverable ()"], {
+            "Asserted": Alert(
+                "fault", "critical",
+                ("Lower Non-Recoverable threshold breached for '{}'. [{}]."
+                "Its outside expected limits"),
+                ("Other System Fans may adjust speed to maintain "
+                 "optimal server temperature."),
+                ("System should recover by itself overtime."
+                 "However, If issue persists for long, contact Seagate support")
+            ),
+            "Deasserted": Alert(
+                "fault_resolved", "informational",
+                "'{} Sensor has recovered from the failure'. [{}]",
+                "None", "None"
+            )
+        }),
+        **dict.fromkeys(["Lower Critical", "Lower Critical ()"], {
+            "Asserted": Alert(
+                "fault", "critical",
+                ("Lower Critical threshold breached for '{}'. [{}]."
+                "Its outside expected limits"),
+                ("Other System Fans may adjust speed to maintain "
+                 "optimal server temperature."),
+                ("System should recover by itself overtime."
+                 "However, If issue persists for long, contact Seagate support")
+            ),
+            "Deasserted": Alert(
+                "fault_resolved", "informational",
+                "'{} Sensor' recovered from the failure. [{}]",
+                "None", "None"
+            )
+        }),
+        **dict.fromkeys(["Lower Non-Critical", "Lower Non-Critical ()"], {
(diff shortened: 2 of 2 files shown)


low-level/sensors/impl/generic/node_hw.py
```diff
@@ -958,19 +959,30 @@ def _parse_fan_info(self, index, date, _time, device_id, sensor_id,\
...
         if status.lower() == "deasserted" and event.lower() == "fully redundant":
             alert_type = "fault"
+            description = f"{sensor_name} sensor is completely redundant"
         elif status.lower() == "asserted" and event.lower() == "fully redundant":
             alert_type = "fault_resolved"
+            description = f"{sensor_name} sensor has recovered from 'fault' state."
         elif "going high" in event.lower() or "going low" in event.lower():
             if status.lower() == "deasserted":
                 alert_type = "fault_resolved"
+                description = f"{sensor_name} sensor is going {state}"
```
sumedhak27, line 976: `System Fan 1A sensor is going high`,  doesn't really tell me what happened to the Fan @gargpalak048.
Also for the Ipmitool events. assert = fault and de-assert = fault_resolved doesn't match exactly for every event,
```
Sensor States:
  lnr : Lower Non-Recoverable
  lcr : Lower Critical
  lnc : Lower Non-Critical
  unc : Upper Non-Critical
  ucr : Upper Critical
  unr : Upper Non-Recoverable
```
I haven't read in-depth, but these states must be representing some threshold values and event is about going low and high wrt those threshold values. 
so going low wrt upper critical value  - Asserted might be a fault_resolved. (I don't know 🤷‍♂️. This might be totally wrong as well. 😅, but we need to check.)

sandeepsgt: You can use `Lower Non-Recoverable threshold breached for {sensor}` if event is lnr, same for other event

gargpalak048: @sumedhak27 so in the specific_info, we are parsing the sensor readings and event occurred.
For eg :-  "event": "Asserted - Lower Non-recoverable going low"

earlier the description was same as the event field. 
so to make it more readable, I simply it

> "description": "System Fan 1A sensor is going low",
"impact": "System Fan 1A is not in good health", 
"recommendation": "Please Contact Seagate Support."

what could be more efficient description? Any suggestions?
description: System Fan 1A sensor is in {} state. ( Lower Non-Recoverable, Lower Critical etc) something like this?

@mariyappanp  @sandeepsgt @cdeshmukh any suggestions

low-level/sensors/impl/generic/node_hw.py
```diff
@@ -958,19 +959,30 @@ def _parse_fan_info(self, index, date, _time, device_id, sensor_id,\
...
+        recommendation = None
+        description = ""
+        state = "high" if "going high" in event.lower() else "low"
         if status.lower() == "deasserted" and event.lower() == "fully redundant":
             alert_type = "fault"
+            description = f"{sensor_name} sensor is completely redundant"
         elif status.lower() == "asserted" and event.lower() == "fully redundant":
             alert_type = "fault_resolved"
+            description = f"{sensor_name} sensor has recovered from 'fault' state."
         elif "going high" in event.lower() or "going low" in event.lower():
```
sumedhak27, line 973: Also rather than having `if-else` conditions to decide description etc. I think we should have some structured hard code as I tried to create for Discrete Sensors like Disk and PSU, we can use the same or have some better structure for Threshold-based sensors since they have the same events for all the threshold-based sensors like temp, voltage, current, etc.

gargpalak048: > Also rather than having if-else conditions to decide description etc.

This code was already in-place, if-else condition is to decide alert type based on event
I only introduced description field

gargpalak048: @sumedhak27  addressed your comment.
I have aligned the changes for Fan module similar to Disk/PSU

low-level/sensors/impl/generic/node_hw_alerts_info.py
```diff
@@ -0,0 +1,116 @@
...
+
+from collections import namedtuple
+from framework.base.sspl_constants import DEFAULT_RECOMMENDATION
+
+Alert = namedtuple('AlertInfo',
+                   ['type', 'severity', 'description',
+                    'impact', 'recommendation'])
+
+alert_for_event = {
+    "Fan": {
```
sumedhak27, line 24: Is it possible to make it generic for all threshold-based senors @gargpalak048 ❓    
The events seem to be the same for all. 
Else looks great to me. 👍

sumedhak27 (approved):

sandeepsgt (approved):

low-level/sensors/impl/generic/node_hw.py
```diff
@@ -48,6 +49,7 @@
 from message_handlers.node_data_msg_handler import NodeDataMsgHandler
 from sensors.INode_hw import INodeHWsensor
 from framework.utils.ipmi_client import IpmiFactory
+from sensors.impl.generic.node_hw_alerts_info import alert_for_event, Alert
```
shriya-deshmukh, line 52: Seems 'Alert' is unused import, can we remove it.

low-level/sensors/impl/generic/node_hw_alerts_info.py
gargpalak048: I kept a tuple per event to make it more loosely bound.
If we need to modify any field for particular event, we could do that.

gargpalak048: But yes, we can do something like,

> **dict.fromkeys(["Sensor threshold event", "Sensor threshold event ()"], {

            "Asserted": Alert(
                "fault", "critical",
                "{event-name} threshold breached for '{}'. [{}]",
                "Fan speed is outside acceptable limits",
                DEFAULT_RECOMMENDATION
            ),
   
then it will be tightly bound to all the events.
What could be the bettter approach?

sumedhak27: No, no. not about events, I was trying to cover all the threshold-based sensors with the same key, so if we can have a single key for all threshold-based sensors (eg. "Threshold-based-sensor") for Fan, temp, voltage, current, etc, instead of having a separate key for each of them.
But not needed now, I think. If such optimization is required we can handle it later.

low-level/sensors/impl/generic/node_hw.py
gargpalak048: yes done

shriya-deshmukh (approved):

low-level/sensors/impl/generic/node_hw.py
```diff
@@ -33,7 +33,8 @@
 from framework.base.internal_msgQ import InternalMsgQ
 from framework.base.module_thread import SensorThread
 from framework.base.sspl_constants import (
-    DATA_PATH, BMCInterface, PRODUCT_FAMILY, ServiceTypes, node_key_id)
+    DATA_PATH, BMCInterface, PRODUCT_FAMILY, ServiceTypes, node_key_id,
+    DEFAULT_RECOMMENDATION)
```
sumedhak27, line 37: `DEFAUTL_RECOMMENDATION` is also not needed now.

gargpalak048: removed unused var

low-level/sensors/impl/generic/node_hw_alerts_info.py
```diff
@@ -0,0 +1,116 @@
...
+                "fault_resolved", "informational",
+                "'{} Sensor has recovered from the failure'. [{}]",
+                "None", "None"
+            )
+        }),
+        **dict.fromkeys(["Lower Critical", "Lower Critical ()"], {
+            "Asserted": Alert(
+                "fault", "critical",
+                "Lower Critical threshold breached for '{}'. [{}]",
+                "Fan speed is outside acceptable limits",
```
cdeshmukh, line 53: "Fan speed is below recommended limits",

low-level/sensors/impl/generic/node_hw_alerts_info.py
```diff
@@ -0,0 +1,116 @@
...
+            "Deasserted": Alert(
+                "fault", "critical",
+                "'{} Sensor', is Fully redundant . [{}]",
+                "Fan efficiency will be impacted",
+                DEFAULT_RECOMMENDATION)
+        }),
+        **dict.fromkeys(["Lower Non-Recoverable", "Lower Non-Recoverable ()"], {
+            "Asserted": Alert(
+                "fault", "critical",
+                "Lower Non-Recoverable threshold breached for '{}'. [{}]",
```
cdeshmukh, line 39: How this description , impact, recommendation helping here exactly ?

Need something like below for every case
description -> Fan {} RPM Lower Non-Recoverable threshold breached, its below acceptable limits.
impact -> Other System Fans may increase speed to maintain optimal server temperature.
recommendation -> System should recover by itself overtime. However, If issue persists for long, contact Seagate support

low-level/sensors/impl/generic/node_hw_alerts_info.py
```diff
@@ -0,0 +1,116 @@
...
+                "fault_resolved", "informational",
+                "'{} Sensor' recovered from the failure. [{}]",
+                "None", "None"
+            )
+        }),
+        **dict.fromkeys(["Lower Non-Critical", "Lower Non-Critical ()"], {
+            "Asserted": Alert(
+                "fault", "critical",
+                "Lower Non-Critical threshold breached for '{}'. [{}]",
+                "Fan speed is outside acceptable limits",
```
cdeshmukh, line 66: "Fan speed is below expected limits",

low-level/sensors/impl/generic/node_hw.py
```diff
@@ -958,19 +959,21 @@ def _parse_fan_info(self, index, date, _time, device_id, sensor_id,\
         original_event = event
         original_status = status
 
-        if status.lower() == "deasserted" and event.lower() == "fully redundant":
```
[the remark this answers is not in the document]
gargpalak048: yes

low-level/sensors/impl/generic/node_hw_alerts_info.py
```diff
@@ -135,5 +135,114 @@
...
+        **dict.fromkeys(["Fully Redundant", "Fully Redundant ()"], {
+            "Asserted": Alert(
+                "fault_resolved", "informational",
+                "'{} Sensor' has recovered from a failure. [{}]",
+                "None", "None"),
+            "Deasserted": Alert(
+                "fault", "critical",
+                "'{} Sensor', is Fully redundant . [{}]",
+                "Fan efficiency will be impacted",
+                DEFAULT_RECOMMENDATION)
```
cdeshmukh: > "'{} Sensor', is Fully redundant . [{}]",

Description here looks doubtful, wrong

Since its fault, i believe it can be
Description -> {} sensor lost redundancy
Impact -> Other System Fans may run with increased speed to maintain optimal temperature
Recommendation -> Rectify Fan issue soon. If not sure, contact Seagate Support, as persistent problem may impact server availability

low-level/sensors/impl/generic/node_hw_alerts_info.py
gargpalak048: > "Fan speed is below expected limits",

Fan speed can be below/above expected limit based on threshold.
Keeping it as -

> Upper Non-Recoverable threshold breached for '{}'. [{}]"
   "Its outside expected limits"

low-level/sensors/impl/generic/node_hw_alerts_info.py
gargpalak048: Ack.
Improvised the field description

low-level/sensors/impl/generic/node_hw_alerts_info.py
gargpalak048: In current code, the description is as 
Deasserted - Fully Redundant
alert_type = "fault"

does it Fan sensor lost redundancy?

gargpalak048: > Description -> {} sensor lost redundancy

{} sensor is running with full redundant speed. It might cause failure

cdeshmukh: >{} sensor is running with full redundant speed. It might cause failure

what is full redundant speed. Its physical fan redundancy we are talking about

gargpalak048: addressed.

cdeshmukh (approved):

cdeshmukh:
test this please

PR #498, opened 2021-08-11
EOS-23183: LDR R2 SSPL: Sw Services monitoring - introduce configurable wait time before raising alert

shriya-deshmukh:
Signed-off-by: Shriya Deshmukh <shriya.deshmukh@seagate.com>

## Problem Statement:

For External services for major state transitions, alerts are raised immediately, HA will take action based on alerts.
But there is a possibility that systemd may be trying to recover the service as per the service file configs already, which would make service active quickly and HA may not need to take any action. 

## Solution Design:

Introduce threshold_wait time before raising alert for external service major state transition.

## Coding:

## Testing:

## Integration:

## PR checklist:
* [ ] Did you add Jira number to the PR?

* [ ] DCO and cla-signed?

### DIFF
--- a/low-level/sensors/impl/centos_7/service_monitor.py
+++ b/low-level/sensors/impl/centos_7/service_monitor.py
@@ -70,7 +70,7 @@ class ActiveState:
     @staticmethod
     def enter(service):
         for set_type in [
-            service.non_active,
+                service.non_active,
                 service.failed_waiting_services,
                 service.active_waiting_services]:
             set_type.discard(service.name)
@@ -85,7 +85,7 @@ class FailedState:
     @staticmethod
     def enter(service):
         for set_type in [
-            service.non_active,
+                service.non_active,
                 service.failed_waiting_services,
                 service.active_waiting_services]:
             set_type.discard(service.name)
@@ -170,7 +170,7 @@ class Service:
     # into the active_waiting_services list.
     active_waiting_services = set()
     # if service state changed to Failed state add service name
-    # in failed_service list.
+    # in failed_waiting_services list.
     failed_waiting_services = set()
     monitoring_disabled = set()
 
--- a/low-level/sensors/impl/centos_7/service_monitor.py
+++ b/low-level/sensors/impl/centos_7/service_monitor.py
@@ -69,8 +69,11 @@ class ActiveState:
 
     @staticmethod
     def enter(service):
-        service.state_enter_timestamp = time.time()
-        Service.non_active.discard(service.name)
+        for set_type in [
+            service.non_active,
+                service.failed_waiting_services,
+                service.active_waiting_services]:
+            set_type.discard(service.name)
 
 
 class FailedState:
@@ -81,8 +84,11 @@ class FailedState:
 
     @staticmethod
     def enter(service):
-        service.state_enter_timestamp = time.time()
-        Service.non_active.discard(service.name)
+        for set_type in [
+            service.non_active,
+                service.failed_waiting_services,
+                service.active_waiting_services]:
+            set_type.discard(service.name)
 
 
 class InactiveState:
@@ -95,6 +101,10 @@ class InactiveState:
     def enter(service):
         service.nonactive_enter_timestamp = time.time()
         Service.non_active.add(service.name)
+        for set_type in [
+                service.failed_waiting_services,
+                service.active_waiting_services]:
+            set_type.discard(service.name)
 
 
 class DisabledState:
@@ -157,11 +167,11 @@ class Service:
     alerts = Queue()
     non_active = set()
     # once fault resolved detected for service, add service name
-    # into the active_services list.
-    active_services = set()
+    # into the active_waiting_services list.
+    active_waiting_services = set()
     # if service state changed to Failed state add service name
     # in failed_service list.
-    failed_services = set()
+    failed_waiting_services = set()
     monitoring_disabled = set()
 
     def __init__(self, unit):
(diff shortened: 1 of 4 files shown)


low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -80,6 +81,7 @@ class FailedState:
 
     @staticmethod
     def enter(service):
+        service.state_enter_timestamp = time.time()
```
gargpalak048, line 84: I am not able to understand why enter() is implemented twice? although it has same functionality

shriya-deshmukh: for each service state we have defined  class(ActiveState,FailedState, InactiveState etc) and for each class we have defined enter() method.
So for eg,  If service state changes to "active",  enter() method from ActiveState class get called and take necessary action.

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -235,15 +240,22 @@ def properties_changed_handler(self, interface, changed_properties,
                     Service.active_services.add(self.name)
                 if self._service_state is InactiveState:
                     self.new_service_state(ActiveState)
+                if self.name in Service.failed_services:
+                    Service.failed_services.discard(self.name)
```
cdeshmukh, line 244: this new_service_state() should take care of discarding service name from other state lists, or its becoming too much repetitive code
If not seen straightforward to address, keep a TODO

sspl_test/alerts/os/test_service_monitor_sensor.py
```diff
@@ -97,7 +97,7 @@ def test_service_failed_alert(args):
     check_sspl_ll_is_running()
     check_service_is_running(service_name)
     simulate_service_alerts.simulate_fault_alert()
-    time.sleep(5)
+    time.sleep(WAIT_TIME)
```
cdeshmukh, line 100: Changing 5 to 50 seconds sleep directly, why ?

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -80,7 +84,11 @@ class FailedState:
 
     @staticmethod
     def enter(service):
-        Service.non_active.discard(service.name)
+        for set_type in [
+            service.non_active,
+                service.failed_waiting_services,
+                service.active_waiting_services]:
```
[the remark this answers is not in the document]
shriya-deshmukh: Done.

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -68,8 +69,11 @@ class ActiveState:
 
     @staticmethod
     def enter(service):
-        service.active_enter_timestamp = time.time()
-        Service.non_active.discard(service.name)
+        for set_type in [
+            service.non_active,
+                service.failed_waiting_services,
+                service.active_waiting_services]:
```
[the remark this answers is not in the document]
shriya-deshmukh: Done.

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -231,19 +246,24 @@ def properties_changed_handler(self, interface, changed_properties,
         if self.state != self.previous_state:
             if self.state == "active":
                 if self._service_state is FailedState:
+                    self.waiting_timestamp = time.time()
+                    Service.active_waiting_services.add(self.name)
```
sumedhak27, line 250: Hi, so Instead of keeping the actions here in if-else condition, what if we create a `ActiveWaiting` & `FailedWaiting` state classes and add these 2 lines in `enter()` method of those classes, then all we have to do will be changing the state to ActiveWaiting here, and in our main loop we will keep checking if waiting time is over to switch to state to Active.

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -231,19 +246,24 @@ def properties_changed_handler(self, interface, changed_properties,
...
+                elif self._service_state is InactiveState:
                     self.new_service_state(ActiveState)
-                    Service.active_services.add(self.name)
-                if self._service_state is InactiveState:
-                    self.new_service_state(ActiveState)
+                # TODO: Handle discarding service from state list
+                # in new_service_state function.
+                Service.failed_waiting_services.discard(self.name)
```
sumedhak27, line 255: this you have already added in `Inactive:enter()` I think.

sumedhak27, line 255: with the above impl, this will be taken care by the respective class enter() method.

low-level/sensors/impl/centos_7/service_monitor.py
```diff
@@ -231,19 +246,24 @@ def properties_changed_handler(self, interface, changed_properties,
...
-                self.new_service_state(FailedState)
+                    self.waiting_timestamp = time.time()
+                    Service.failed_waiting_services.add(self.name)
+                Service.active_waiting_services.discard(self.name)
             else:
                 # Avoid raising duplicate alerts
                 if self._service_state != FailedState:
                     self.new_service_state(InactiveState)
+                Service.active_waiting_services.discard(self.name)
+                Service.failed_waiting_services.discard(self.name)
```
sumedhak27, line 266: similar way if we add `InactiveWaiting` state then actions under if-else will reduce,
*_But after adding these states we will again have to calculate transitions and verify actions so **I am good with the current changes**._

low-level/sensors/impl/centos_7/service_monitor.py
sumedhak27: *_But after adding these states we will again have to calculate transitions and verify actions so **I am good with the current changes**._

low-level/sensors/impl/centos_7/service_monitor.py
sumedhak27: *_After adding these states we will again have to calculate transitions and verify actions so **I am good with the current changes**._

sumedhak27 (approved):

cdeshmukh (approved):

sandeepsgt (approved):

mariyappanp (approved):

PR #518, opened 2021-08-25
EOS-23324: nw & nw:cable alert Improvement

sumedhak27:
## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?

### DIFF
--- a/low-level/message_handlers/node_data_msg_handler.py
+++ b/low-level/message_handlers/node_data_msg_handler.py
@@ -91,10 +91,7 @@ class NodeDataMsgHandler(ScheduledModuleThread, InternalMsgQ):
         'disk': 0
     }
     prev_nw_status = {}
-    prev_cable_cnxns = {}
-    # Dir to maintain fault detected state for interface
-    # in case of cable fault detection
-    interface_fault_state = {}
+    prev_cable_status = {}
     FAULT = "fault"
     FAULT_RESOLVED = "fault_resolved"
 
@@ -211,10 +208,8 @@ class NodeDataMsgHandler(ScheduledModuleThread, InternalMsgQ):
             if resource == 'nw':
                 self.prev_nw_status = \
                     self.persistent_data[resource].get('prev_nw_status', {})
-                self.prev_cable_cnxns \
-                    = self.persistent_data[resource].get('prev_cable_cnxns', {})
-                self.interface_fault_state \
-                    = self.persistent_data[resource].get('interface_fault_state', {})
+                self.prev_cable_status = \
+                    self.persistent_data[resource].get('prev_cable_status', {})
             elif self.persistent_data[resource]\
                 [f'high_{resource}_usage'].lower() == "true":
                 self.high_usage[resource] = True
@@ -228,9 +223,8 @@ class NodeDataMsgHandler(ScheduledModuleThread, InternalMsgQ):
                             f'{data_path}_{self.node_id}')
         if resource == 'nw':
             self.persistent_data[resource] = {
-                'prev_nw_status' : self.prev_nw_status,
-                'prev_cable_cnxns' : self.prev_cable_cnxns,
-                'interface_fault_state' : self.interface_fault_state
+                'prev_nw_status': self.prev_nw_status,
+                'prev_cable_status': self.prev_cable_status
             }
         else:
             self.persistent_data[resource] = {
@@ -720,20 +714,21 @@ class NodeDataMsgHandler(ScheduledModuleThread, InternalMsgQ):
                     # Store the state to Persistent Cache.
                     self.persist_state_data('cpu', 'CPU_USAGE_DATA')
 
-    def _send_ifdata_json_msg(self, sensor_type, resource_id, resource_type, state, severity, event=""):
+    def _send_ifdata_json_msg(self, resource_id, resource_type, state, event=""):
         """A resuable method for transmitting IFDataMsg to RMQ and IEM logging"""
         ifDataMsg = IFdataMsg(self._node_sensor.host_id,
                                 self._node_sensor.local_time,
                                 self._node_sensor.if_data,
                                 resource_id,
                                 resource_type,
-                                state, severity, event)
+                                state,
+                                self.severity_reader.map_severity(state), event)
         # Add in uuid if it was present in the json request
         if self._uuid is not None:
             ifDataMsg.set_uuid(self._uuid)
         jsonMsg = ifDataMsg.getJson()
         self.if_sensor_data = jsonMsg
-        self.os_sensor_type[sensor_type] = self.if_sensor_data
+        self.os_sensor_type["nw"] = self.if_sensor_data
 
         # Transmit it to message processor
         self._write_internal_msgQ(EgressProcessor.name(), jsonMsg)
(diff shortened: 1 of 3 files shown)


low-level/message_handlers/node_data_msg_handler.py
```diff
@@ -742,144 +737,66 @@ def _send_ifdata_json_msg(self, sensor_type, resource_id, resource_type, state,
...
+            else: # no nw_fault
+                if self.prev_nw_status.get(if_name) == "nw_down":
+                    # raise nw_fault_resolved_alert
+                    self.prev_nw_status[if_name] = "nw_up"
+                    self._send_ifdata_json_msg(
+                        if_name, self.NW_RESOURCE_TYPE, self.FAULT_RESOLVED,
+                        nw_alert_desc.format(if_name, "UP"))
+                if self.prev_cable_status.get(if_name) == "cable_disconnected":
+                    # raise nw_cable_fault_resolved_alert
+                    self.prev_cable_status[if_name] = "cable_connected"
```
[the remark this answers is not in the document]
sumedhak27: Agreed @mariyappanp, but have a question about what action to take in the following scenario.
```python
if nw_up:
    if cable_connected and prev_cable status == "DISCONNECTED":
        raise cable_FAULT_RESOLVED
    elif prev_cable status == "DISCONNECTED":
        # what action to take here when the previous alert is nw_disconnected and we didn't send the nw_down alert,
        # and now that the nw is up still cable is disconnected, what should be the action ??
    
    if prev_nw_status == "DOWN":
        raise nw_FAULT_RESOLVED
```

low-level/message_handlers/node_data_msg_handler.py
```diff
@@ -720,20 +714,21 @@ def _generate_cpu_data(self):
...
+    def _send_ifdata_json_msg(self, resource_id, resource_type, state, event=""):
         """A resuable method for transmitting IFDataMsg to RMQ and IEM logging"""
         ifDataMsg = IFdataMsg(self._node_sensor.host_id,
                                 self._node_sensor.local_time,
                                 self._node_sensor.if_data,
                                 resource_id,
                                 resource_type,
-                                state, severity, event)
+                                state,
+                                self.severity_reader.map_severity(state), event)
```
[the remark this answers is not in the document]
mariyappanp: I'm fine, however each alert will carry its own information in specific_info.

low-level/message_handlers/node_data_msg_handler.py
```diff
@@ -742,144 +737,66 @@ def _send_ifdata_json_msg(self, sensor_type, resource_id, resource_type, state,
...
+            if_name = interface["ifId"]
+            nw_status = interface['nwStatus'].upper()
+            # if nw_fault
+            if nw_status in ["DOWN", "UNKNOWN"]:
+                nw_cable_status = interface['nwCableConnStatus'].upper()
+                # if nw_cable_disconneted
+                if nw_cable_status == "DISCONNECTED":
+                    if self.prev_cable_status.get(if_name) != "cable_disconnected":
+                        # raise nw_cable_fault_alert
+                        self.prev_cable_status[if_name] = "cable_disconnected"
```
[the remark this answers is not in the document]
mariyappanp: Current status is fine, but this need to use actual, right? `self.prev_cable_status[if_name] = "cable_disconnected"`

low-level/message_handlers/node_data_msg_handler.py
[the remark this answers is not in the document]
mariyappanp:
```
if nw_up:
    if cable_connected and prev_cable status == "DISCONNECTED":
        raise cable_FAULT_RESOLVED
    elif not cable_connected and prev_cable status == "CONNECTED":
        # We can ignore prev_cable status 'UNKNOWN' as we have nw_up
        raise cable_FAULT

    if prev_nw_status == "DOWN":
        raise nw_FAULT_RESOLVED
````
Any other gap?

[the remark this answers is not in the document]
Comment (sumedhak27) [reply low-level/message_handlers/node_data_msg_handler.py]: Addressed 👍 . No gaps now, just the thing is we are giving alert of cable_disconnected and connected even if nw_status is UP.

File: low-level/message_handlers/node_data_msg_handler.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (sumedhak27) [reply low-level/message_handlers/node_data_msg_handler.py]: sorry, not getting you😓. can you explain again?

**Review by mariyappanp — APPROVED** [review]

**Review by sandeepsgt — APPROVED** [review]

**Review by shriya-deshmukh — APPROVED** [review]

File: low-level/message_handlers/node_data_msg_handler.py [hunk 4]
```diff
@@ -720,20 +714,21 @@ def _generate_cpu_data(self):
...
-    def _send_ifdata_json_msg(self, sensor_type, resource_id, resource_type, state, severity, event=""):
+    def _send_ifdata_json_msg(self, resource_id, resource_type, state, event=""):
         """A resuable method for transmitting IFDataMsg to RMQ and IEM logging"""
-        ifDataMsg = IFdataMsg(self._node_sensor.host_id,
-                                self._node_sensor.local_time,
-                                self._node_sensor.if_data,
-                                resource_id,
-                                resource_type,
-                                state, severity, event)
+        if_data = [inf for inf in self._node_sensor.if_data \
```
cdeshmukh, line 719: So till now all nw interfaces data were getting packed in, instead of the one for which alert is generated ? This is better approach then

low-level/message_handlers/node_data_msg_handler.py
```diff
@@ -742,144 +737,75 @@ def _send_ifdata_json_msg(self, sensor_type, resource_id, resource_type, state,
...
+                        # raise nw_fault_alert
+                        self.prev_nw_if_status[if_name] = "DOWN"
+                        self._send_ifdata_json_msg(
+                            if_name, self.NW_RESOURCE_TYPE, self.FAULT,
+                            nw_alert_desc.format(if_name, "DOWN"))
+                # nw_cable_state is unknown and nw_is_faulty
+                else:
+                    if self.prev_nw_if_status.get(if_name) != "DOWN":
+                        # raise nw_fault_alert
+                        self.prev_nw_if_status[if_name] = "DOWN"
```
cdeshmukh, line 782: DOWN or UNKNOWN

low-level/sensors/impl/generic/node_data.py
```diff
@@ -250,27 +252,14 @@ def _fetch_nw_status(self):
     def fetch_nw_cable_conn_status(self, interface):
         carrier_status = None
         try:
-            carrier_status = self._utility_instance.fetch_nw_cable_status(self.nw_interface_path, interface)
+            carrier_status = Network().get_link_state(interface)
+        except NetworkError as err:
+            logger.debug(err)
+            carrier_status = "UNKNOWN"
         except Exception as e:
-            if e == errno.ENOENT:
```
cdeshmukh, line 255: why we wont need these specific exceptions now

low-level/sensors/impl/generic/node_data.py
```diff
@@ -250,27 +252,14 @@ def _fetch_nw_status(self):
...
-                    "Problem occured while reading from nw carrier file:"
-                    f" {self.nw_interface_path}/{interface}/carrier."
-                    "Operation is not permitted.")
-            else:
-                logger.error(
-                    "Problem occured while reading from nw carrier file:"
-                    f" {self.nw_interface_path}/{interface}/carrier. Error: {e}")
+            logger.excpetion(
+                "Problem occured while reading from nw carrier file:"
+                f" {self.nw_interface_path}/{interface}/carrier. Error: {e}")
```
cdeshmukh, line 262: this exception error string covers the error already ?

sumedhak27: Yes Chetan, In NetworkError, the logic is for all the OSError Exceptions, so this error log is for any other Exception that is not OSError, for that we are simply printing the Exception error string.

low-level/message_handlers/node_data_msg_handler.py
sumedhak27: `prev_nw_if_status` is value set by us to check if the prev status and for caching purpose it is set to "DOWN"/"UP" only @cdeshmukh , so UNKNOWN for nw_interface_status will be considered as "DOWN" and the variable will have "DOWN" value.

low-level/sensors/impl/generic/node_data.py
```diff
@@ -250,27 +252,14 @@ def _fetch_nw_status(self):
     def fetch_nw_cable_conn_status(self, interface):
         carrier_status = None
         try:
-            carrier_status = self._utility_instance.fetch_nw_cable_status(self.nw_interface_path, interface)
+            carrier_status = Network().get_link_state(interface)
+        except NetworkError as err:
+            logger.debug(err)
```
cdeshmukh, line 257: this should be logger.error only

low-level/sensors/impl/generic/node_data.py
cdeshmukh: > In NetworkError, the logic is for all the OSError Exceptions

I dont see any OSError being handled in NetworkError !?
https://github.com/Seagate/cortx-monitor/blob/main/low-level/framework/platforms/server/error.py#L39

low-level/sensors/impl/generic/node_data.py
sumedhak27: That might not be the error per se, 
if some interface is disabled, then it won't have a carrier file so the Exception will be NetowrkError since OS related errors are handled by NetworkError, In this case we don't want to keep logging error every iteration, this is the expected UNKNOWN state. That's why I logged in debug. If the Exception is unknown or unexpected then only I am logging it.

low-level/sensors/impl/generic/node_data.py
sumedhak27: while raising the NetworkError.

low-level/sensors/impl/generic/node_data.py
cdeshmukh: Add these details in code then, why such info is not shared until asked ? These are not obvious things we can relate just by looking at the code being changed or even if someone looking at just this or limited code

low-level/sensors/impl/generic/node_data.py
cdeshmukh: Can you elaborate ?
NetworkError class code reference i gave, there i dont see any OSError being handled, how its happening then

sumedhak27:
```python
## /platforms/server/network.py:get_link_state
except OSError as err:
        raise NetworkError(err.errno, (
            "Failed to read link state for interface '%s' "
            "due to an Error: '%s, %s'"),
            interface, err.strerror, err.filename)
````
We are handling all the OSErrors in the same way. These are not errors for us but a known state of UNKNOWN
Separate error handling of each type of OSError is not required,
OS Errors Handled in previous code, 
1. File Does not exist =>  We don't want to raise error, but status is set to UNKNOWN
2. Not enough permission => will never occur /carrier is read-only file no one can change the file permission
3. Operation not permitted => reading the file is the only operation and it is always supported for /carrier file.

low-level/sensors/impl/generic/node_data.py
sumedhak27: Added in the comments please check.

low-level/sensors/impl/generic/node_data.py
cdeshmukh: Ok i can relate now

low-level/sensors/impl/generic/node_data.py
cdeshmukh: Its not crisp, but thats fine.

cdeshmukh (approved):

PR #521, opened 2021-08-31
EOS-23221: Replace SSPL Logger with Cortx Logger class

gargpalak048:
Signed-off-by: Palak Garg <palak.garg@seagate.com>

## Problem Statement:

SSPL component should use CORTX logger instead of implementing its own logger class

## Solution Design:

Moving to common CORTX Logger class for SSPL component

## Coding:

## Testing:

## Integration:

## PR checklist:

* [x] DCO and cla-signed?


low-level/framework/utils/service_logging.py
```diff
@@ -19,130 +19,16 @@
...
 
-import logging.handlers
-import time
-import os
+from cortx.utils.log import Log
 
+def init_logging(dcs_service_name, file_path, log_level="INFO"):
+    """Initialize logging for SSPL component"""
+    Log.init(service_name=dcs_service_name, log_path=file_path, level=log_level)
+    logger.info("Logging has been initialized for SSPL !!!")
```
sumedhak27, line 27: maybe ..
```python
logger.info(f"Logging has been initialized for {dcs_service_name} !!!")
```

low-level/framework/utils/webservices.py
```diff
@@ -23,7 +23,8 @@
 
 import requests
 from requests.exceptions import Timeout, ConnectionError, HTTPError
-from framework.utils.service_logging import logger
+from framework.base.sspl_constants import SSPL_LOG_PATH
+from cortx.utils.log import Log as logger
```
sumedhak27, line 27: Why here we can't use `from framework.utils.service_logging import logger` ??

low-level/framework/utils/service_logging.py
gargpalak048: done

low-level/framework/utils/webservices.py
gargpalak048: done.
Note:- we need to initialize logger obj in webservices.py file as well because realstor login api is called even before sspl_ll_d script.

low-level/framework/utils/service_logging.py
```diff
@@ -19,130 +19,23 @@
...
+import sys
 import os
+from cortx.utils.log import Log
 
+def init_logging(dcs_service_name, file_path, log_level="INFO"):
+    """Initialize logging for SSPL component."""
+    try:
+        Log.init(service_name=dcs_service_name, log_path=file_path, level=log_level)
+    except Exception as err:
+        print("[ Error ] when initializing logging :")
```
cdeshmukh, line 31: Push these error statements to system syslog in case of exceptions. Also optimize the statement
>"[ Error ] CORTX Logger Init failed with error {err}"

low-level/framework/utils/service_logging.py
```diff
@@ -19,130 +19,23 @@
...
+        print(err)
+        print("Exiting ...")
+        sys.exit(os.EX_USAGE)
```
cdeshmukh, line 34: Also why exiting with EX_USAGE, when it may not be usage related error ! EX_SOFTWARE  seems proper

>	EX_USAGE -- The command was used incorrectly, e.g., with
 \*		the wrong number of arguments, a bad flag, a bad
 \*		syntax in a parameter, or whatever.

> 	EX_SOFTWARE -- An internal software error has been detected.
 \*		This should be limited to non-operating system related
 \*		errors as possible.

low-level/framework/utils/service_logging.py
```diff
@@ -19,130 +19,23 @@
...
-import logging.handlers
-import time
+import sys
 import os
+from cortx.utils.log import Log
 
+def init_logging(dcs_service_name, file_path, log_level="INFO"):
+    """Initialize logging for SSPL component."""
+    try:
+        Log.init(service_name=dcs_service_name, log_path=file_path, level=log_level)
```
cdeshmukh, line 29: what about log rotation config ? What are the defaults, if any exists, does it server SSPL needs ?

low-level/framework/utils/utility.py
```diff
@@ -152,14 +152,15 @@ def replace_expr(filename:str, key, new_str:str):
                 f.write(line)
 
     @staticmethod
-    def set_ownership_recursively(root_dir, uid, gid):
+    def set_ownership_recursively(list_root_dir, uid, gid):
```
cdeshmukh, line 155: check that no existing functions are affected

low-level/framework/utils/webservices.py
```diff
@@ -47,6 +48,7 @@ class WebServices(object):
     def __init__(self):
         super(WebServices, self).__init__()
 
+        init_logging("sspl", SSPL_LOG_PATH)
```
cdeshmukh, line 51: why separate init here again?

sumedhak27: reply from @gargpalak048, 
```smalltalk
Note:- we need to initialize logger obj in webservices.py file as well because realstor login api is called even before sspl_ll_d script.
```

low-level/framework/utils/service_logging.py
```diff
@@ -19,130 +19,23 @@
  ****************************************************************************
 """
 
-import logging.handlers
-import time
+import sys
 import os
+from cortx.utils.log import Log
 
+def init_logging(dcs_service_name, file_path, log_level="INFO"):
```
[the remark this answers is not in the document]
gargpalak048: this was the old code, not sure about dcs term.
Any change required?

low-level/framework/utils/service_logging.py
gargpalak048: Done

low-level/framework/utils/service_logging.py
gargpalak048: same comment. This was the old code,  just moved it to service_logging file from sspl_ll_d.
But Agree, EX_SOFTWARE makes more sense,
optimized the code

low-level/framework/utils/service_logging.py
gargpalak048: Log rotation is handled by cortx-utils itself.
Default config in cortx-utils is,

> backup_count=10,
file_size_in_mb=10,
max_bytes = file_size_in_mb * 1024 * 1024

most of the components like HA using this default config only, I think for SSPL also this will suffice

low-level/framework/utils/utility.py
gargpalak048: yes verified

low-level/framework/utils/service_logging.py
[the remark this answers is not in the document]
cdeshmukh: oh ok legacy DCS, there used to be another module along with SSPL daemon. we can safely rename that in new code, just keep service_name

low-level/framework/utils/service_logging.py
cdeshmukh: yes it suffice. but still add comments about default config around this code and also mention its recommended to refer to latest code for confirming config. that may help in future

low-level/framework/utils/webservices.py
cdeshmukh: How? sspl-ll_d has main() which starts the sspl daemon. You mean webservice class used before logging init? Then that init should simply happen first

jjxsg: When [RealStorEnclosure](https://github.com/Seagate/cortx-monitor/blob/main/low-level/framework/platforms/realstor/realstor_enclosure.py) is imported, its `init` method is called in which we have a `login()` function which performs login through `realstor` API. So even before SSPL starts, we have a call to `realstor` API that uses functions from webservice class.

sumedhak27 (approved):

cdeshmukh:
thanks @gargpalak048, as you know @jjxsg took care of this through separate PR #523 (given he was having permission issue to update this same PR), while you were sick out. Am just mentioning here for reference

PR #523, opened 2021-09-06
EOS-23221: Replace SSPL Logger with CORTX Logger Class

jjxsg:
## Problem Statement:

SSPL should use CORTX logger instead of implementing its own logger class.
## Solution Design:

Use CORTX logger and remove SSPL's own logging implementation.
## Coding:

## Testing:

## Integration:

* [ ] If there is any interface change, other component gate gatekeepers are ready to accept the change?

## PR checklist:

* [x] DCO and cla-signed?


low-level/framework/utils/service_logging.py
```diff
@@ -19,130 +19,33 @@
...
 
-import logging.handlers
-import time
+import sys
 import os
+import syslog
+from cortx.utils.log import Log
+
+
+def init_logging(service_name, file_path, log_level="INFO"):
```
jjxsg, line 28: > **cdeshmukh**: legacy DCS, there used to be another module along with SSPL daemon. we can safely rename that in new code, just keep service_name

Done

low-level/framework/utils/service_logging.py
```diff
@@ -19,130 +19,33 @@
...
+    """Initialize logging for SSPL component."""
+    # Log rotation is configured within cortx-utils with
```
jjxsg, line 30: > **cdeshmukh**: what about log rotation config? What are the defaults, if any exists, does it server SSPL needs?
> 
> **gargpalak048:** Log rotation is handled by cortx-utils itself.
> Default config in cortx-utils is,
> backup_count=10,
> file_size_in_mb=10,
> max_bytes = file_size_in_mb * 1024 * 1024
> most of the components like HA using this default config only, I think for SSPL also this will suffice
> 
> **cdeshmukh**: yes that suffice. but still add comments about default config around this code and also mention its recommended to refer to latest code for confirming config. that may help in future

Done

low-level/framework/utils/webservices.py
```diff
@@ -47,6 +48,7 @@ class WebServices(object):
     def __init__(self):
         super(WebServices, self).__init__()
 
+        init_logging("sspl", SSPL_LOG_PATH)
```
jjxsg, line 51: > **cdeshmukh:** why separate init here again?
> 
> **gargpalak048:** We need to initialize logger obj in webservices.py file as well because realstor login api is called even before sspl_ll_d script.
>  
> **cdeshmukh:** How? sspl-ll_d has main() which starts the sspl daemon. You mean webservice class used before logging init? Then that init should simply happen first
> 

When [RealStorEnclosure](https://github.com/Seagate/cortx-monitor/blob/main/low-level/framework/platforms/realstor/realstor_enclosure.py) is imported, its `init` method is called in which we have a call to `login()` function which performs login through `realstor` API. So even before SSPL starts, we have a call to `realstor` API that uses functions from webservice class.

low-level/files/etc/rsyslog.d/0-iemfwd.conf
```diff
@@ -1,30 +0,0 @@
-# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
sandeepsgt, line 1: This file is needed for journal to file IEM forwarding, IEM sensor will not work without this.
This file is not related to SSPL logs.

low-level/framework/utils/webservices.py
cdeshmukh: hmm so in effect two logging handles created for same sspl log file !? We will have to fix this through separate patch, will not drag this one for that

low-level/files/etc/rsyslog.d/0-iemfwd.conf
cdeshmukh: Ok, thanks for identifying. we should keep it then

low-level/files/etc/logrotate.d/iem_messages
```diff
@@ -1,14 +0,0 @@
-/var/log/cortx/iem/iem_messages
```
cdeshmukh, line 1: as @sandeepsgt identified, we will need IEM logging (& logrotate) related config still, not related to SSPL logging

low-level/files/etc/rsyslog.d/0-iemfwd.conf
jjxsg: Addressed, please check.

low-level/files/etc/logrotate.d/iem_messages
jjxsg: Addressed, please check.

cdeshmukh (approved):

sumedhak27 (approved):

low-level/framework/utils/filestore.py
```diff
@@ -117,15 +119,15 @@ def _load_json_file(self, key):
             except:
                 value = fh.read()
         except IOError as err:
-            logger.warning("I/O error[{0}] while loading data from file {1}): {2}"\
+            logger.warn("I/O error[{0}] while loading data from file {1}): {2}"\
```
satish-seagate, line 122: why we changed the warning to warn?
As I understand logging.warn has been deprecated since Python 3.3
https://stackoverflow.com/a/15655674

jjxsg: Yes, `warn` is depreciated in native python logger. But the logger object here is coming from a custom CORTX Log class which has `warn` method and not `warning`. 
Ref: https://github.com/Seagate/cortx-utils/blob/main/py-utils/src/utils/log.py

satish-seagate: Okay

satish-seagate (approved):

low-level/framework/utils/service_logging.py
mariyappanp: @jjxsg , While trying out with this change, logs are suppressed in sspl.log also same is writing into /var/log/messages.

